PR #34, opened 2026-08-28
Repair the thinking round-trip instead of killing the session

### LINKED ISSUE #32: Cross-family failover replays another provider's thinking blocks, so one 502 kills the session with a 400 at attempt 1/18
**Version:** kitty-bridge 1.5.0 (pip), Python 3.12.14, Claude CLI 2.1.238, Ubuntu 24.04 self-hosted GitHub Actions runner
**Consumer:** Claude Code Action, headless, `--json-schema` structured output, `--effort max`
**Profile:** MiniMax pool primary (`minimax-large`/`-mid`/`-small`), API-key DeepSeek backend as backup. `ANTHROPIC_MODEL=deepseek-v4-flash`

## Summary

When a mid-conversation failover crosses **provider families**, the bridge replays the accumulated transcript to the new backend unchanged. The transcript's assistant turns were produced by the *previous* provider, so a receiving backend running in thinking mode rejects the request:

```
400 {"error":{"message":"The `content[].thinking` in the thinking mode must be passed back to the API.",
     "type":"invalid_request_error","param":null,"code":"invalid_request_error"}}
```

Because 400 is not retryable, the bridge surfaces it to the client and the session dies — **on failover attempt 1 of 18, with 17 attempts still available.**

One transient 502 from the primary therefore kills an otherwise-healthy session. In our case: 48 successful requests, 125 seconds, **$4.56 of tokens**, discarded. The consumer is a required CI check, so the pull request goes red with no review.

Failover *within* one family works fine — the same profile recovered from two consecutive 429s in an earlier run. **Crossing the family boundary is the trigger.**

## What we observed

Two runs of the same workflow on the same PR, one day apart.

### Run B — cross-family failover, killed the session

`Overrode model` counts for the whole session:

```
$ grep -oE "Overrode model: [A-Za-z0-9.-]+ -> [A-Za-z0-9.-]+" timeline.txt | sort | uniq -c
     49 Overrode model: deepseek-v4-flash -> MiniMax-M3
      1 Overrode model: MiniMax-M3 -> deepseek-v4-flash
```

48 requests served by MiniMax-M3, all 200. Then:

```
14:22:10.013 DEBUG Upstream POST → https://api.minimax.io/anthropic/v1/messages
14:22:10.230 DEBUG Upstream response status: 502
14:22:10.231 INFO  Backend minimax-large marked unhealthy for 300s after hard error
14:22:10.231 DEBUG Overrode model: MiniMax-M3 -> deepseek-v4-flash
14:22:10.231 INFO  Messages stream failover: attempt 1/18 (status 502), switching backend
14:22:11.253 DEBUG Upstream response status: 400
14:22:11.254 INFO  Backend deepseek-1 marked unhealthy for 300s after hard error
14:22:11.254 ERROR Upstream error 400: {"error":{"message":"The `content[].thinking` in the thinking mode
                   must be passed back to the API.", ...}}
14:22:11.494 INFO  Bridge server stopped
14:22:11.495 INFO  Bridge process exiting.
```

Claude Code's execution record:

```json
"terminal_reason": "api_error",
"api_error_status": 400,
"result": "API Error: 400 {\"error\":{\"message\":\"The `content[].thinking` in the thinking mode must be passed back to the API.\", ...}}",
"duration_ms": 125175,
"modelUsage": { "deepseek-v4-flash": { "inputTokens": 227060, "outputTokens": 9762,
                "cacheReadInputTokens": 6356529, "costUSD": 4.557614499999999 } }
```

### Run A — same-family failover, recovered cleanly

Same profile, same consumer, the day before. Two failovers, both inside the MiniMax pool:

```
22:54:54.876 DEBUG Upstream response status: 429
22:54:54.877 INFO  Backend minimax-mid marked unhealthy for 300s after rate_limit error
22:54:54.877 INFO  Messages stream failover: attempt 1/18 (status 429), switching backend
22:54:55.820 DEBUG Upstream response status: 429
22:54:55.821 INFO  Backend minimax-small marked unhealthy for 300s after rate_limit error
22:54:55.821 INFO  Messages stream failover: attempt 2/18 (status 429), switching backend
22:54:57.148 DEBUG Upstream response status: 200      ← recovered
```

The transcript replayed without complaint because the receiving backend was another MiniMax. That run died later for an unrelated reason (`structured_output_retry_exhausted`); it is included only as the control showing failover itself is not broken.

## Why it happens

`server.py` normalizes assistant `thinking` blocks on the **response** path only — `server.py:2750` turns a `thinking` block into `content_block_start` / `thinking_delta` SSE events for the client. There is no corresponding handling on the **request** path: the client's messages array is forwarded as-is to whichever backend the failover lands on.

So after a cross-family switch the new backend receives assistant turns it did not author, and a backend that requires thinking blocks echoed back in thinking mode has no valid way to accept them.

The failure is then made terminal by the retry classifier:

```python
_RETRYABLE_STATUSES = {429, 500, 502, 503, 504}   # server.py:398
```

`_should_retry_stream(400, body)` returns `False` — correctly, for a *genuine* upstream 400. But this 400 was caused by the bridge's own request construction, not by anything the client asked for and not by the backend being unwell. Two consequences:

1. **The failover budget is abandoned at attempt 1/18.** A sibling MiniMax backend would very likely have accepted the same transcript, as Run A demonstrates.
2. **`deepseek-1` is marked unhealthy for 300 s** for a request the bridge malformed. That is a healthy backend taken out of rotation for five minutes, and it will misroute subsequent sessions.

## Suggested fixes, in the order we would want them

1. **Prefer a same-provider backend when failing over mid-conversation.** Cross-family only once same-family options are exhausted. This alone would have saved both the session and the cost, and needs no transcript rewriting.
2. **Normalize the transcript when failover does cross families** — strip assistant `thinking` blocks (and any provider-specific `signature`) that the target cannot accept, or re-emit them in its expected shape.
3. **Do not attribute a post-switch 400 to the backend's health.** A request-shape rejection arriving on the first request after a backend switch is evidence about the bridge's translation, not about the backend. Continue the failover budget instead of surfacing, and leave the backend's health untouched.

Fix 3 is worth having even with 1 and 2, because it converts this class of bug from "session dies" into "session continues, log shows a compatibility miss".

## Open question

In Run B the bridge chose DeepSeek even though Run A shows `minimax-mid` and `minimax-small` exist in the profile, and neither is recorded as unhealthy anywhere in Run B's log. Is backend selection on failover ordered/weighted in a way that skipped the siblings, or were they unavailable for a reason the log does not show? If a sibling was healthy and simply not preferred, fix 1 is the whole fix.

## Reproduction

Not deterministic — it needs a real upstream 502/429 on the primary while a cross-family backup is configured. It should be reproducible by pointing a profile at a primary that returns 502 and a DeepSeek backup, running any thinking-mode session long enough to accumulate assistant turns, then forcing the primary to fail.

Happy to supply the full 79 MB bridge debug log, the execution record, or the filtered timelines from either run.

shelpuk:
Fixes #32.

A user's CI job died mid-run and lost $4.56 of work for no good reason.

Kitty can be configured with a pool of model providers, so that if one has a
hiccup the session quietly moves to another. That worked — except when the
replacement came from a *different vendor*.

Every turn, the coding agent re-sends the whole conversation so far. Some
vendors (DeepSeek, Kimi) require the model's own reasoning to be handed back
with that history. A conversation that started on MiniMax has no such reasoning
attached, so the moment Kitty moved the session to DeepSeek, DeepSeek rejected
the whole conversation as malformed.

Kitty treated that rejection as "this provider is broken": it gave up on the
session, showed the user a raw API error, and benched a perfectly healthy
provider for five minutes. In the reported incident that threw away 48
successful requests and 125 seconds of work **on the first of eighteen
available retries**, and the customer's pull request went red with no review.

After this change, Kitty recognises that the rejection is about the *request it
built*, not the provider. It fixes up the conversation, retries the same
provider, and the session carries on. The user sees nothing but a line in the
log. One vendor hiccup no longer costs a session.

## What changed

When an upstream returns a 4xx saying thinking-mode reasoning must be passed
back, the bridge now:

1. recognises it as a **request-shape rejection**, not a backend fault;
2. gives every assistant turn the empty carrier its target requires — a
   `thinking` block on the Anthropic wire, `reasoning_content` on the Chat
   Completions wire;
3. retries the **same backend**: no health penalty, no failover slot spent;
4. remembers the backend for the session, so later turns are repaired before
   their first attempt instead of paying a rejected round-trip each time.

If the repair changes nothing, today's mark-unhealthy/failover path runs
unchanged — the change can't make any existing case worse.

## Two design points worth a reviewer's attention

**The repair targets the serialized upstream body, never `cc_request`.** This is
load-bearing twice:

- A later failover re-serializes a clean transcript, so a sibling backend never
  receives a carrier fabricated for a different provider. Anthropic requires
  passed-back `thinking` blocks to be complete, unmodified and
  signature-verified, and to be *stripped* when switching models — shipping an
  empty one to a sibling could turn a recoverable failover into a second 400,
  charged to that sibling's health.
- The repair sees carriers the provider adapter already injected during
  serialization (`kimi`, `zai`, `custom_openai` add `reasoning_content` inside
  `translate_to_upstream`), so it reports "no change" rather than retrying
  byte-identical bytes.

**The dialect comes from the adapter, not the request flag.** New
`ProviderAdapter.upstream_wire_is_messages_api` (True on `AnthropicAdapter` and
its subclasses). The `tool_use` format fallback clears
`_native_messages_request` while an Anthropic-wire adapter still emits an
Anthropic body via the inherited CC → Messages translation — branching on the
flag would write `reasoning_content` onto an Anthropic-shaped message.

## Why the issue's "fix 1" is not implemented

The issue proposed preferring a same-provider backend on failover, and asked
whether the MiniMax siblings were healthy but simply not preferred. They were
not: the session died at attempt 1 of 18, so the budget was not exhausted, which
means no backend was healthy after DeepSeek was marked.

More durably, `_get_backend_family` derives family from `provider_type`, which
is `custom_anthropic` for **both** MiniMax and DeepSeek in an Anthropic-shaped
pool. The boundary that matters is the upstream *host*, which nothing models.
Recorded in `SYSTEM_DESIGN.md` §8.B.9 so it doesn't read as an oversight.

## Testing

40 new tests in `tests/bridge/test_thinking_roundtrip_{repair,failover}.py` —
L1 for the predicate and the rewriter, L3 against a scripted fake upstream for
the retry, health, stickiness, per-backend binding and fall-through behaviour.

Each regression test was falsified: reverting the corresponding fix turns it red
with a legible message, including the headline test, which fails with the
issue's own 400-vs-200 symptom.

| Gate | Result |
|---|---|
| `tests/bridge/` | 1071 passed, 1 skipped |
| `tests/` minus bridge & egress-proxy | 1854 passed, 35 skipped |
| `ruff check .` | clean |
| `mypy src/kitty` | clean |
| `lint-imports` | 3 contracts kept |

`tests/test_egress_https_proxy.py` has 5 fixture errors that are pre-existing —
confirmed by reproducing them on a clean tree. They need a local TLS proxy.

## One open assumption

That an **empty** carrier satisfies these providers is supported by DeepSeek's
docs and by third-party repro (claude-code-router#1378 reports a direct curl
with `reasoning_content: ""` succeeding), but is **not verified by us against a
live key** — the incident isn't reproducible without one. For Kimi the external
fixes use `" "` instead, while this repo's own
`ProviderAdapter._inject_empty_reasoning_content` has shipped `""` in
production; the repair uses `""` to match the helper it sits beside.

The design degrades safely either way: a still-rejected repaired request falls
through to today's behaviour, so a wrong assumption costs one attempt and cannot
make the session worse than it is now. Symptom to watch for is a backend that
never converges — the same rejection every turn.

https://claude.ai/code/session_01NJJJmUgnet1Qdh2imTzdSL

### DIFF
--- a/src/kitty/bridge/server.py
+++ b/src/kitty/bridge/server.py
@@ -314,6 +314,60 @@ def _is_thinking_roundtrip_error(status: int, body: object) -> bool:
     return any(all(token in searchable for token in tokens) for tokens in _THINKING_ROUNDTRIP_PATTERNS)
 
 
+def _with_thinking_carrier(msg: dict, *, native: bool) -> dict | None:
+    """Return a copy of an assistant message carrying the thinking carrier.
+
+    Adds the empty carrier the target's thinking mode requires: a leading
+    ``{"type": "thinking", "thinking": ""}`` content block on the Anthropic
+    wire, a ``reasoning_content`` field on the Chat Completions wire.  The
+    block goes first because the Messages API requires reasoning to precede
+    text in an assistant turn.
+
+    Returns ``None`` — meaning "already satisfied, nothing to do" — for a
+    message that carries a thinking block or a ``reasoning_content`` already,
+    and for one whose ``content`` is neither a list nor a string, which is not
+    a shape the carrier can be attached to.  That is what makes the repair
+    converge: the caller retries only on a change, so a message reported as
+    changed on every pass would retry forever.
+
+    On the Chat Completions side the test is ``is None`` rather than the
+    ``not in`` used by :meth:`ProviderAdapter._inject_empty_reasoning_content`.
+    The two are deliberately different: that helper runs proactively during
+    serialization, while this one runs after it and must treat an explicit
+    ``None`` as still needing repair, yet leave an already-written ``""``
+    alone.  Do not unify them — convergence depends on this asymmetry.
+
+    Args:
+        msg: The assistant message to inspect. Never modified.
+        native: True when the enclosing body is an Anthropic Messages body,
+            False when it is a Chat Completions body.
+
+    Returns:
+        A new message dict with the carrier added, or None when the message
+        already satisfies the contract or cannot carry one.
+    """
+    if not native:
+        if msg.get("reasoning_content") is not None:
+            return None
+        return {**msg, "reasoning_content": ""}
+
+    content = msg.get("content")
+    if isinstance(content, list):
+        if any(isinstance(block, dict) and block.get("type") == "thinking" for block in content):
+            return None
+        return {**msg, "content": [dict(_THINKING_CARRIER_BLOCK), *content]}
+
+    if isinstance(content, str):
+        # A string-content turn needs the block form to carry thinking; the
+        # text survives as an explicit text block.
+        blocks: list[dict] = [dict(_THINKING_CARRIER_BLOCK)]
+        if content:
+            blocks.append({"type": "text", "text": content})
+        return {**msg, "content": blocks}
+
+    return None
+
+
 def _repair_thinking_roundtrip(body: dict, *, native: bool) -> bool:
     """Give every assistant turn the thinking carrier its target requires.
 
@@ -336,12 +390,16 @@ def _repair_thinking_roundtrip(body: dict, *, native: bool) -> bool:
 
     The dialect is passed in rather than inferred, because
     ``{"role": "assistant", "content": "..."}`` is valid in both and guessing
-    would silently write the wrong carrier for one of them.
+    would silently write the wrong carrier for one of them.  It must come from
+    the adapter's :attr:`ProviderAdapter.upstream_wire_is_messages_api`, not
+    from ``_native_messages_request`` — see :meth:`BridgeServer._upstream_body_for`.
 
     Changed messages are **copied**, not edited in place, and ``messages`` is
     replaced with a new list.  ``translate_to_upstream`` returns a shallow copy
     whose ``messages`` list is ``cc_request``'s own, so editing a message would
     reach back into the request that the next attempt re-serializes.
+
+    Per-message rewriting lives in :func:`_with_thinking_carrier`.
 
     Args:
         body: The outgoing upstream body.  Its ``messages`` key is rebound when
@@ -362,39 +420,12 @@ def _repair_thinking_roundtrip(body: dict, *, native: bool) -> bool:
     repaired = messages.copy()
     changed = False
     for index, msg in enumerate(messages):
-        # The contract binds assistant turns only; user, system and tool
-        # messages carry no reasoning to pass back.
+        # The contract binds assistant turns only.
         if not isinstance(msg, dict) or msg.get("role") != "assistant":
             continue
-
-        if not native:
-            # Chat Completions: a missing sibling field is the gap.  Tested for
-            # absence rather than falsiness so an already-repaired ``""`` is not
-            # reported as a change — the caller retries on a change, and a
-            # repair that never converges would retry forever.  Deliberately
-            # `is None` rather than the `not in` used by the sibling helper
-            # ``ProviderAdapter._inject_empty_reasoning_content``, which runs
-            # proactively during serialization: an explicit ``None`` reaches
-            # this path already serialized and still needs repairing.  Do not
-            # "unify" the two — the convergence guarantee depends on this one.
-            if msg.get("reasoning_content") is None:
-                repaired[index] = {**msg, "reasoning_content": ""}
-                changed = True
-            continue
-
-        content = msg.get("content")
-        if isinstance(content, list):
-            if any(isinstance(b, dict) and b.get("type") == "thinking" for b in content):
-                continue
-            repaired[index] = {**msg, "content": [dict(_THINKING_CARRIER_BLOCK), *content]}
-            changed = True
-        elif isinstance(content, str):
-            # A string-content turn still needs the block form to carry
-            # thinking; the text survives as an explicit text block.
-            blocks: list[dict] = [dict(_THINKING_CARRIER_BLOCK)]
-            if content:
-                blocks.append({"type": "text", "text": content})
-            repaired[index] = {**msg, "content": blocks}
+        carrier = _with_thinking_carrier(msg, native=native)
+        if carrier is not None:
+            repaired[index] = carrier
             changed = True
 
     if changed:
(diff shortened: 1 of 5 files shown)


shelpuk:
## Self-review pass — what I changed and what I'm deferring

Pushed `070ca73`. All 8 checks were green before it and it is behaviour-preserving
(the same 15 tests kill a mutant that disables the native carrier, before and
after the refactor).

### Changed

- **Extracted `_with_thinking_carrier`** from `_repair_thinking_roundtrip`, so
  the loop reads as bookkeeping and each dialect's rule stands alone. The
  `is None` vs `not in` asymmetry with
  `ProviderAdapter._inject_empty_reasoning_content` now has one documented home
  rather than being an inline aside — that asymmetry is what makes the repair
  converge, so it was the wrong thing to leave buried.
- **Trimmed two oversized in-body comment blocks** (18 and 9 lines). The
  rationale already lives in the docstrings and `SYSTEM_DESIGN.md` §8.B.9.
- **Three convergence-guard tests** for shapes the carrier cannot attach to.
  The one I'd draw a reviewer's eye to: a bridge server can register *all four*
  protocol routes (`server.py:1928-1939`), so a backend in the sticky set may
  serialize a **Gemini** body, which has `contents` and no `messages`. That is
  a no-op today, but it was true by luck rather than by test.

### Deferred, with reasons

**1. The empty carrier is still unverified against a live key.** DeepSeek's docs
plus a third-party curl repro support `""`; for Kimi the external fixes use
`" "` while this repo's own `_inject_empty_reasoning_content` has shipped `""`
in production. I kept `""` to match the helper it sits beside so the two cannot
drift. Deferred because it needs a key I don't have. **Symptom if wrong:** a
backend that never converges — same rejection every turn, one wasted round-trip
each. Bounded, and it cannot make a session worse than today's behaviour.

**2. The non-streaming Messages path does not detect the rejection.** It still
marks the backend unhealthy on this 400. Claude Code streams every real turn, so
the reported failure cannot arrive there; adding detection means threading
repair-and-retry through a second retry loop for a case we have not observed.
Recorded as FR-8 so the narrower scope reads as a decision.

**3. The sticky set is not conditioned on the current turn having thinking
enabled.** A later non-thinking turn still gets the carrier. Detecting "thinking
is off" reliably across both dialects would add a second inference path, and if
a provider does object the cost is one 400 falling through to normal failover.
Now documented in §8.B.9 with the trigger to revisit.

**4. A repair landing on the last pre-`_EMPTY_FINAL_DELAYS` attempt** continues
into that branch and logs "Empty upstream response: final retry", which is
misleading. Pre-existing and shared with the adjacent `tool_use` branch —
changing it here would either diverge the two or touch unrelated behaviour.

**5. `.system_design/steps/` does not exist in this repo**, so the per-feature
step artifact has no home. Flagged rather than silently starting a convention.

Happy to pick up any of these in this PR if you'd rather not defer them.

Issue #33, opened 2026-08-28
MiniMax /anthropic: intermittent envelope-wrapped tool_use inputs — invisible in kitty logs, no failover


mudrykliliya:
## Summary

Our Claude Code review CI (routed through kitty 1.5.0) hits intermittent structured-output failures: the CLI's final `StructuredOutput` tool call — validated by ajv against a `--json-schema` schema — is rejected at the **root** on every one of its 5 attempts. All serving traffic in the failing pool is **MiniMax-M3 via `api.minimax.io/anthropic`** (standard transport), so the question for kitty is whether anything at the MiniMax adapter layer can detect, log, or mitigate this.

## Setup

- kitty 1.5.0 (unpinned from PyPI at install time)
- Balancing pool: 3× MiniMax-M3 profiles at `api.minimax.io/anthropic` (`minimax` provider, standard transport) + 1 DeepSeek backup member
- Client: Claude Code CLI with `--json-schema`; the schema is compiled into an ajv validator behind a `StructuredOutput` tool
- Review session shape: a 62+ turn tool chain ending in one `StructuredOutput` call

## What we can prove from our own logs (evidence tier 1)

- All 5 structured-output attempts rejected with the same ajv triple on every observed failing run:
  - `root: must NOT have additional properties`
  - `must have required property 'findings'`
  - `must have required property 'conversation_notes'`
- Read together: the payload had one extra root-level key and was missing the real fields — the fingerprint of the actual content nested one level too deep inside a wrapper key.
- Intermittent on identical input: same commit, same prompt, same schema — two runs failed, two passed minutes apart (controlled experiment, 2026-08-27/28).
- We have **not** captured the wrapper key from our own bytes yet. A CLI-side capture step is landing in our review workflow that records every `StructuredOutput` attempt's input from the action's execution record; we will post the captured payload here on the next failing run.

## External finding 1: MiniMax's `/anthropic` endpoint wraps tool arguments

*(evidence tier 2 — externally documented, not observed in our own bytes)*

[Langertha](https://metacpan.org/pod/Langertha::Engine::MiniMaxAnthropic) (a Perl LLM client library) hit this exact defect on this exact endpoint and documented it. Verbatim from their canonical doc:

[...]

PR #35, opened 2026-08-28
Make response-side tool_use observable, and flag inputs that contradict their schema

shelpuk:
Addresses #33 — asks 1 and 2. **Not** ask 3 (the defensive unwrap); see below.

A customer's CI review job kept failing, and nobody could tell whose fault it was.

Kitty sits between the coding agent and the model provider. When the model
calls a tool, it hands back a small structured payload. Sometimes a provider
returns that payload in the wrong shape — the agent rejects it and the run
fails. Everything still looks healthy from the outside: the provider answered
"200 OK", so kitty counted it as a success.

The customer could not investigate, because kitty wrote nothing about what it
had passed along. They instrumented an entire run to capture it and got **zero
matching lines in a 222,551-line debug log** — the logging existed on only one
of five internal code paths, and not the one they were using.

Now every tool payload kitty returns is recorded, on every path. When a payload
contradicts the shape the agent itself asked for, kitty says so by name — which
tool, which fields, and which provider in the pool produced it — and counts it
where an operator will see it without re-running anything.

Kitty does **not** try to repair the payload. We do not yet know for certain
what is malformed about it, and a wrong guess would quietly corrupt a correct
answer instead of loudly failing. This change is what will tell us.

## What changed

- Every `tool_use` returned to a Messages client is logged at DEBUG with its
  name and a truncated input — on **all five** routes: native passthrough,
  CC-translated and custom-transport streaming, plus both non-streaming paths.
- When an input contradicts that tool's own declared `input_schema`, a WARNING
  names the tool, the offending keys and the serving backend.
- A per-backend `malformed_tool_use` counter in `GET /stats` and the shutdown
  summary.
- Every line carries the prefix `tool_use audit:`, so the next incident is one
  `grep` away.

## Three decisions worth a reviewer's attention

**Observed at the write boundary.** All streaming paths end by writing
Anthropic SSE downstream, so watching there gives one parser and one state
machine, needs no translator changes, and never audits an abandoned upstream
attempt — those events are never written. The interrupted-stream finalizations
write downstream too, so they are audited and `finish()`ed like any other exit.
That one is not a corner case: #33 reports the failure as *intermittent on
identical input*, and an upstream truncating after emitting the tool call is
the archetype — exactly when the counter must not read zero.

**The rule is narrow on purpose.** It fires on both an unexpected root key AND
a missing required property, or on a single root key whose value holds every
missing required name. A root `required` absent from the input is *proof* of a
violation regardless of `additionalProperties` — that half cannot
false-positive. A model merely omitting a field, or adding a harmless extra
key, stays silent. It stands down entirely for composed schemas: draft-07,
which the Claude Code SDK validates with, ignores `$ref` siblings, so a valid
input would otherwise trip both conditions. A detector that fires on ordinary
traffic trains people to ignore it, which costs more than it gives.

**It counts as well as logs.** `kitty/__init__.py` attaches only a
`NullHandler`, and the only real handler is the `--debug` file handler — so a
WARNING alone would reach nobody in a normal run, which is precisely the
complaint the issue makes.

## Safety

Diagnostics only. Nothing rewrites a payload or touches health, cooldown or
routing. A leaked exception would be severe — `_is_retryable_exception` returns
False for `ValueError`/`TypeError`/`KeyError`, so it would escape the attempt
loop, inject an SSE `error` into a stream that had already delivered valid
content, and leave the backend unmarked. So assembler and detector are both
contained, the auditor disables itself after any containment event rather than
failing once per chunk, all three of its buffers are bounded, and a fresh
auditor is built per upstream attempt so a failover cannot splice two attempts
into one fabricated block.

**The WARNING carries key names and lengths, never values.** It is the only
line that can reach an embedder's own handler, and tool inputs routinely carry
file contents and user data. Pinned by a test asserting a secret value never
appears.

## Why no unwrap (ask 3)

The issue gates it on a capture the reporter states they have **not** made, and
notes it "has a chance of refuting the envelope hypothesis entirely". An unwrap
is not a safe bet on that: a tool whose input legitimately is a single `result`
key — plausible for one named `StructuredOutput` — would be silently corrupted,
turning a loud validator rejection into quiet data loss. A missed detection
costs one more failing run; a wrong unwrap costs a wrong answer nobody sees.

The detector is deliberately schema-driven and names no cause, so it stays
correct and useful whichever way the capture lands — and it is what will settle
it. If the wrapper key is confirmed, the unwrap becomes a small, evidenced
follow-up.

I also verified the Langertha citation in the issue is accurate verbatim, and
established by experiment that `api.minimax.io/anthropic` traffic can only run
the **native passthrough** path (the translated path yields zero events for
Anthropic-shaped SSE) — so the native-streaming test is the acceptance test for
this incident.

## Testing

58 new tests across three files — L1 detector, L1 SSE assembler, L3 subsystem.
The precision rule, the detector, and the truncated-stream fix were each
mutation-checked: reverting them turns the relevant tests red.

| Gate | Result |
|---|---|
| `tests/bridge/` | 1132 passed, 1 skipped |
| `tests/` minus bridge & egress-proxy | 1854 passed, 35 skipped |
| `ruff` / `mypy` / `lint-imports` | clean |

`tests/test_egress_https_proxy.py` has 5 pre-existing fixture errors, unrelated
(they need a local TLS proxy). One existing contract test in
`test_session_stats.py` needed the new key added — it pins the exact stats
shape, which is it doing its job.

## Scope

Messages protocol only. The Responses, Gemini and Chat Completions paths carry
tool calls too; widening is a follow-up.

https://claude.ai/code/session_01NJJJmUgnet1Qdh2imTzdSL

### DIFF
--- a/README.md
+++ b/README.md
@@ -355,6 +355,12 @@ their token totals, how many times a request switched backend (`failovers`), and
 (`all_backends_unhealthy`). Like `/healthz`, it is readable without credentials unless a keys file is configured — and
 `kitty claude` deliberately runs without one.
 
+`malformed_tool_use` counts responses where an upstream returned `200 OK` with a `tool_use` whose `input` contradicts
+the schema the client declared for that tool — the shape a client-side validator then rejects. It is reported per
+backend and as a session total, so a pool member returning well-formed-looking garbage is visible without re-running
+under `--debug`. It is a diagnostic only: it never marks a backend unhealthy or triggers failover. Run with `--debug`
+and `grep 'tool_use audit:'` for the offending payloads.
+
 **After the run.** `--session-summary PATH` (or `KITTY_SESSION_SUMMARY`) writes the same document to a file when the
 bridge shuts down — a small artifact CI can upload, instead of a multi-megabyte debug log:
 
--- a/src/kitty/bridge/server.py
+++ b/src/kitty/bridge/server.py
@@ -41,6 +41,7 @@ from kitty.bridge.responses.events import (
     format_error_event as responses_format_error,
 )
 from kitty.bridge.responses.translator import ResponsesTranslator
+from kitty.bridge.tool_audit import AUDIT_MARKER, ToolUseAuditor, collect_tool_schemas, report_tool_use
 from kitty.cloudflare import is_cloudflare_block
 from kitty.egress import EgressConfig, should_bypass
 from kitty.providers.base import ProviderAdapter, ProviderError
@@ -978,6 +979,12 @@ class BridgeServer:
         self._stats_retries = 0
         self._stats_all_unhealthy = 0
         self._stats_backend_attempts: dict[int, int] = {}
+        # Malformed tool_use responses per backend index (issue #33).  Counted
+        # rather than only logged because the WARNING reaches a handler only
+        # under --debug, and the point of the ask is to distinguish "200 but
+        # garbage" from "200 and clean" in an ordinary run.  Counting only —
+        # it never influences health, cooldown or routing.
+        self._stats_malformed_tool_use: dict[int, int] = {}
         self._stats_models: dict[str, dict[str, int]] = {}
         self._started_at: str | None = None
 
@@ -1621,6 +1628,7 @@ class BridgeServer:
                         "healthy": health["healthy"],
                         "remaining_cooldown": remaining,
                         "cooldown_events": health.get("failure_count", 0),
+                        "malformed_tool_use": self._stats_malformed_tool_use.get(idx, 0),
                     }
                 )
         else:
@@ -1636,6 +1644,7 @@ class BridgeServer:
                     "healthy": True,
                     "remaining_cooldown": 0,
                     "cooldown_events": 0,
+                    "malformed_tool_use": self._stats_malformed_tool_use.get(-1, 0),
                 }
             )
         return {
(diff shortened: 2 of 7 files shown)


shelpuk:
## Self-review pass — `9c783f7`

All 8 checks were green before this and no reviews were open, so this pass was
me re-reading the code I'd just added. It found one real bug and a fair amount
of waste. Behaviour reported to the operator is unchanged.

### The bug

The line-length bound was checked **before** draining, so a single chunk
carrying >10 MB of *complete* SSE lines would disable the auditor — and then
silently miss every tool call in the rest of that response. The bound exists
for a line that never terminates, so it now applies to the residual after
draining. Pinned by a test that fails against the old ordering.

Rewriting the drain then exposed a second defect, and the **existing** test
caught it: `_handle` can disable the auditor mid-chunk, and my new loop carried
on processing lines afterwards, re-populating the state `_disable` had just
cleared. Worth spelling out why the old code was fine — it stopped only
*accidentally*, because `_disable` rebound `self._line_buffer` and the `while`
condition re-read it each pass. That is precisely the sort of implicit coupling
that breaks under a refactor, so the loop now checks `_disabled` explicitly and
`_disable` clears in place rather than rebinding.

### The waste

Three of them, all on a path that previously did no parsing at all:

- `report_tool_use` serialised the entire tool input, then handed it to
  `logger.debug` — which discards it whenever the debug log is off, i.e. almost
  always. Now behind `isEnabledFor`.
- Every line was UTF-8 decoded *before* the `data:` prefix check, so the
  comment calling it a "cheap pre-check" was doing the expensive half first.
  Now matched on bytes.
- Every `data:` event was JSON-parsed even on turns with no tool call. An event
  can only matter if a block is already open or the raw line contains the
  literal `tool_use`; a substring scan settles that far more cheaply.

Measured with the debug log off:

| Stream | Before | After |
|---|---|---|
| 48 KB tool payload | 1.03 ms | **0.37 ms** |
| 108 KB tool-free text | 2.79 ms | **1.16 ms** |

The drain also copied the whole buffer per line (`bytes(buffer).partition`),
quadratic across a response; it now scans with `find()` and truncates once.

§8.B.10's cost paragraph previously asserted this was cheap without evidence —
it now carries the measurements.

### Clarity

The open-block state was an untyped dict indexed by string keys (`"name"`,
`"args"`, `"size"`, `"seeded"`). It is now an `_OpenBlock` dataclass, so the
fields are named once and checked by mypy.

### One thing I deliberately left as-is

The `if self._disabled: return` guard in the drain loop is **not** covered by a
test — I mutated it away and all 26 auditor tests stayed green, because
`_disable` empties the buffer the loop is scanning, which terminates it anyway.
I kept the guard rather than delete it: relying on that coupling is exactly what
broke a moment earlier. The comment now says it is belt-and-braces and why,
instead of implying it is load-bearing. Flagging it so nobody mistakes an
untested line for an oversight.

### Gates

bridge 527 + 607 passed / 1 skipped · non-bridge 1854 passed / 35 skipped ·
ruff, mypy, lint-imports clean. 60 audit tests; the buffer-bound fix and the
truncated-stream fix were each falsified against the old code.

PR #40, opened 2026-08-31
Never charge a network interruption to a healthy backend (fixes #38); release 1.8.0

### LINKED ISSUE #38: Transient connection issues result in a persistent sequence of 503 errors
Here is the analysis from Claude: 

The failure sequence is right there at 16:48 today. Here's what went wrong:

Root cause: client disconnect, misread as backend outage

16:48:09  STREAM START msg_f88257... → POST z.ai → 200 OK   ← upstream was fine
16:48:25  Backend zai_coding marked unhealthy for 300s after transport error
          (ClientConnectionResetError)
16:48:25  Failover → secondary backend → also marked unhealthy (300s)
16:48:27  ERROR: Upstream POST failed after 10 attempts: Cannot write to closing transport
          Client disconnected before stream EOF
16:48:45  ERROR: All 2 backends unhealthy → 503 Retry-After=280s
17:06:14  Cooldowns expired, backends retried
17:22:21  (last log entry) healthy stream, completed normally

What actually happened: the client side dropped the connection mid-stream at 16:48 (ClientConnectionResetError — the terminal session on the other end went away: window closed, Ctrl+C, or a network blip). z.ai had accepted the request and was streaming normally. But the bridge classified the dead pipe as a backend failure and quarantined both backends for 300 seconds. Any request arriving in that ~5-minute window got an instant 503 "All backends unhealthy" — even though neither provider was actually down. Both recovered on their own at 17:06, and the bridge is healthy as of the last entry (17:22).

shelpuk:
Fixes #38.

Two things could make kitty stop serving even though every provider behind it was healthy.

**Your agent hanging up was blamed on the provider.** If you closed the terminal, hit Ctrl+C, or your laptop's network blipped while an answer was streaming, kitty read that as "the provider dropped the connection". It put that plan in a five-minute cooldown, switched to the next plan, tried to write the same answer to the same dead terminal, and cooled that one down too. Within seconds every plan was quarantined and the next thing you typed got `503 All backends unhealthy` for five minutes — with nothing actually wrong upstream. That is exactly the incident in #38: two healthy z.ai plans knocked out at 16:48 by one disconnect, back to normal at 17:06 only because the cooldowns expired.

**A one-second network hiccup cost you a plan for five minutes.** kitty had no tolerance at all for a transient connection failure to a provider: the first dropped connection was treated as an outage.

After this change:

- A disconnect on your side is charged to your side. The stream ends, every plan stays healthy, and your next message is served normally.
- A dropped connection to a provider gets 30 seconds of benefit of the doubt. kitty quietly retries the *same* plan with escalating backoff; you never see an error for a blip that clears. Only a connection that keeps failing past 30 seconds counts as a real failure and moves you to another plan.
- If a provider drops the connection *after* it has already sent you part of an answer, kitty now closes that answer off cleanly instead of showing an internal error — and does not restart it on another plan, which would have duplicated text you had already seen.

The 30-second tolerance applies on every protocol kitty speaks: Claude Code (Messages), Codex (Responses), Gemini, and OpenAI-compatible chat, streaming and non-streaming.

Released as **1.8.0**.

## What changed technically

Root cause: aiohttp reports a dead *client* with the same `ConnectionResetError` family it uses for a dead *upstream*. `_stream_messages` wrote to the client inside its per-attempt `try`, so a client disconnect landed in the failover handler, `_is_transport_error` said "transport", and `_mark_backend_unhealthy(..., failure_kind="transport")` gave it a 300 s cooldown. The failover repeated the write on the next backend and quarantined that too.

- `_write_client()` wraps every previously-unguarded client write in the Messages retry loop and re-raises as `ClientDisconnectedError`; a dedicated `except` clause ends the stream without touching backend health. `_is_retryable_exception` / `_is_transport_error` both reject the new type, so its message cannot be mistaken for an upstream reset.
- `_ensure_prepared()` assigns `sr` only after `prepare()` returns. `prepare()` writes headers to the client and can fail, so this makes `sr is not None` mean exactly "the client has received something" — the invariant the two new gates below depend on.
- `TransportGrace` is a per-request budget: up to `_TRANSPORT_GRACE_PERIOD` (30 s) across `_TRANSPORT_GRACE_DELAYS` (2/4/8/16 s), bounded by both count and wall clock. Used in the Messages loop, in the new `_open_upstream_stream()` that the Responses / Gemini / Chat Completions handlers open through, and in `_make_upstream_request` (rebuilt as a `while` loop so connection blips and status retries have separate budgets).
- Grace retries are free: the Messages loop iterates `raw_attempt` and derives `attempt = raw_attempt - transport_grace.retries`. Without this a single-backend profile spent its whole allowance on grace and then sat through `_EMPTY_FINAL_DELAYS` (20 s + 40 s) mislabelled as empty-response retries.
- `aiohttp.ClientPayloadError` (truncated upstream body) is now classified as a connection failure, and `asyncio.TimeoutError` is now excluded from `_is_transport_error` — aiohttp files `ServerTimeoutError` under `ClientConnectionError`, so the function had been contradicting its own docstring and would have added 30 s of grace on top of a 120 s `sock_read` timeout.
- An upstream drop after bytes reached the client finalizes the partial message rather than restarting it, matching the existing FI-8.3 truncation policy. The native-passthrough path has no half-open translator message, so it gets a terminal `error` event instead of a stream that simply stops.

## Testing

New `tests/bridge/test_client_disconnect_health.py` — 28 tests, ~1 s. Unit tests for the classification and the grace arithmetic; subsystem tests driving a real `BridgeServer` with `aioresponses` upstreams, plus a local aiohttp server that aborts a socket mid-body for the cases that need a real disconnect.

Each behaviour was verified to fail before its fix, not just to pass after:

- against the pre-fix `server.py`, the four client-disconnect tests fail with the same `Upstream POST failed after 10 attempts: Cannot write to closing transport` line as the production log in #38;
- with the grace delays emptied, all five tolerance tests fail;
- with the `raw_attempt` accounting reverted, the budget test fails on `Empty upstream response: final retry in 20.0s (5/6)`;
- with the native terminal-event fallback removed, the native abort test fails.

Full suite: **3029 passed, 1 failed, 34 skipped**. The failure is `test_bridge_management.py::TestBridgeReachable::test_connect_target_table[::ffff:0.0.0.0-127.0.0.1]`, which is pre-existing — verified failing on the base commit `ffd5d45` and untouched by this branch. `ruff`, `mypy`, and `lint-imports` are clean.

## Known gap

The Responses, Gemini and Chat Completions *streaming* handlers still conflate a dead client and a dead upstream in a single handler-level `except (ConnectionResetError, BrokenPipeError, OSError)` that ends the stream silently. That pre-dates this change; those handlers gain the connect-time grace but not the attribution fix, because giving them one means wrapping three ~250-line loop bodies in a per-attempt handler. Happy to do that as a follow-up if you want it in this release.

https://claude.ai/code/session_015uiTmSTM2JRCizCEEF9DSf

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,6 +222,13 @@ automatically without restarting anything.
 Anything that makes a member unavailable counts: rate limits, exhausted quotas, expired credentials, upstream 5xx, and
 connection failures.
 
+A dropped connection is given the benefit of the doubt first. For up to 30 seconds kitty retries the *same* member
+instead of cooling it down, so a brief network interruption between kitty and a provider costs a short pause rather
+than five minutes of that plan. Only a connection that keeps failing past that window counts as a failure. The same
+applies in the other direction: if your agent disconnects mid-answer — you close the terminal, hit Ctrl+C, or your
+network blips — that is charged to the agent, never to the provider, so the next request still finds every plan
+healthy.
+
 One caveat: a member that cannot handle streaming is skipped for streaming requests regardless of tier, so a streaming
 request can reach a backup member while a non-streaming primary is still healthy. This only affects providers without
 streaming support, and matches how balanced profiles already behaved.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "kitty-bridge"
-version = "1.7.0"
+version = "1.8.0"
 description = "Kitty Bridge — launch coding agents through a local API bridge"
 readme = "README.md"
 license = "MIT"
--- a/src/kitty/__init__.py
+++ b/src/kitty/__init__.py
@@ -4,5 +4,5 @@ import logging
 
 logging.getLogger(__name__).addHandler(logging.NullHandler())
 
-__version__ = "1.7.0"
+__version__ = "1.8.0"
 __all__ = ["__version__"]
--- a/src/kitty/bridge/server.py
+++ b/src/kitty/bridge/server.py
@@ -636,6 +636,14 @@ _STREAM_READ_TIMEOUT = 120  # seconds — upstream must respond with first byte 
 _SINGLE_BACKEND_COOLDOWN_CAP = 30  # seconds — cap cooldown for single-backend profiles
 _CLOUDFLARE_FIRST_HIT_COOLDOWN = 15  # seconds — short cooldown for first Cloudflare block
 _ALL_UNHEALTHY_FAST_FAIL_THRESHOLD = 60  # seconds — fast-fail if soonest retry exceeds this
+# A blip on the wire between kitty and the provider is not a provider outage.
+# Before this window existed, the first connection reset cost the backend a full
+# cooldown, and the failover spread the damage to its siblings until every
+# provider sat in quarantine and healthy upstreams answered 503 (issue #38).
+# Inside the window the same backend is simply retried, so a few seconds of bad
+# Wi-Fi or a dropped keep-alive costs a short pause instead of five minutes.
+_TRANSPORT_GRACE_PERIOD = 30.0  # seconds of connection trouble tolerated per request
+_TRANSPORT_GRACE_DELAYS = (2.0, 4.0, 8.0, 16.0)  # backoff between grace retries; sums to the window
 _SSE_MAX_LINE_BYTES = 10 * 1024 * 1024  # 10 MiB — guards against unbounded line_buffer growth
 _MAX_HEADER_VALUE_CHARS = 256  # cap on X-Kitty-* values — names are config-supplied, so unbounded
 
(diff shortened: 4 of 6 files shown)


shelpuk:
## Self-review pass (3de1298)

I went back over my own diff before asking anyone else to read it. Two real defects fell out, plus some duplication worth collapsing. All CI checks were green on b8276c7 (3.10–3.13); this pushes on top.

### Bugs found and fixed

**The grace window was per upstream *call*, not per request.** `_request_with_retry` calls `_make_upstream_request` once per backend, so a non-streaming request across three backends could have spent the full 30 s tolerance three times over — contradicting the "30 seconds per request" that both the docs and the log message claim, and adding up to ~90 s of sleeping to a request that should have failed over quickly. One `TransportGrace` is now created in `_request_with_retry` and threaded through both retry helpers into `_make_upstream_request`. The helpers now *require* the argument rather than defaulting it, since nothing else calls them and the default was dead code.

**Extending the Messages loop could push it past its attempt budget.** To hand grace retries back I widened the loop by `len(_TRANSPORT_GRACE_DELAYS)`. But `_EMPTY_FINAL_DELAYS` is indexed by `attempt`, and there is exactly one `continue` in that loop which does *not* check `attempt` — the native `tool_use` format fallback. If it fired on the last attempt, the next iteration would index `_EMPTY_FINAL_DELAYS[2]` on a two-element list and raise `IndexError`. The loop now breaks once `attempt` reaches `max_attempts`, restoring the pre-change invariant.

On that second one, being straight about the evidence: the trigger needs a native-passthrough provider to hit a `tool_use` format mismatch on precisely the final attempt. I judged the guard — which just restores the bound that existed before my change — a better use of effort than constructing that scenario, so it ships with an attempt-budget invariant test rather than a live reproduction of the crash.

### Refactors, no behaviour change

- The log-sleep-retry decision existed in three near-identical copies. It is now `_wait_out_transport_blip()`, so the budget is spent in exactly one place and callers only read its boolean answer. In `_stream_messages` the `sr is None` test moved *ahead* of it, so a drop that happened after the client already saw bytes no longer spends grace it cannot use.
- `TransportGrace` read its period at construction but its delays at call time. Both resolve at first use now, so patching either behaves the same — the kind of asymmetry that is invisible until it bites someone.

### Test fixes

- The finish-events test now asserts that some writes actually succeeded before the failing one. Its `fail_after=4` is tied to the translator's event count, so without that assertion a change upstream could silently degrade it into a duplicate of the fail-on-first-write case.
- The empty-response-budget test now reads the log instead of monkeypatching `asyncio.sleep` process-wide, which was a flake waiting to happen.
- New: a per-request-window test (fails if the window is rebuilt per backend) and an attempt-budget invariant test.

### One observation, not fixed

`test_concurrent_backend_selection.py::TestProviderConfigContextIsolation::test_concurrent_tasks_have_isolated_provider_config` is flaky by construction — it asserts two tasks picked *different* backends, but `_select_backend` picks at random from two equally-weighted members, so it fails whenever both land on the same one. Measured over 20 runs: 9/20 failures on this branch, 7/20 on the base commit `ffd5d45` — statistically the same coin flip, and my diff touches neither `_select_backend` nor the `_active_provider_config` contextvar. Leaving it alone as unrelated to this PR; flagging it because it will redden CI at random. Happy to fix it separately if you want.

The other local failure, `test_bridge_management.py::TestBridgeReachable::test_connect_target_table[::ffff:0.0.0.0-127.0.0.1]`, is environment-dependent: it fails on my machine on the base commit too, and passes on all four Python versions in CI.

shelpuk:
## CI fix (9dd2745)

All four test jobs failed on `3de1298`, every one of them on the same assertion:

```
FAILED tests/bridge/test_concurrent_backend_selection.py::TestProviderConfigContextIsolation::test_concurrent_tasks_have_isolated_provider_config
  AssertionError: assert 'https://backend-a.example.com' != 'https://backend-a.example.com'
```

Otherwise green: 3031 passed, 34 skipped, and lint / import-layering / type-check all passed on every version.

This is the flake I flagged in my previous comment, now actually blocking, so I've fixed it rather than left it.

**What was wrong.** The test asserted that two concurrent tasks saw different `provider_config`s, but it never made them *select* different backends — it just hoped `_get_next_backend`, which picks at random, would hand out one of each. With two equally-weighted backends both tasks land on the same one about half the time. The failure message says it plainly: both tasks got `backend-a`.

**The fix.** Selection is pinned so the first caller gets backend "a" and the second gets "b", and each task now asserts it still sees *its own* config after the await. Which backend a task gets is incidental to this test; that the two do not leak into each other across an `await` is the point, and the selection algorithm itself is covered by the other tests in the file.

**Verification.** 30/30 runs pass locally, up from ~50%. More importantly, I checked I hadn't just turned it into an always-pass: replacing the `_active_provider_config` contextvar read with shared instance state — the bug this test exists to catch — still makes it fail.

This is a pre-existing flake, unrelated to the client-disconnect work; it just happened to lose the coin flip on all four jobs at once.

PR #43, opened 2026-09-06
Add a four-layer test suite design (KBR-2)

shelpuk:
Closes [KBR-2](https://shelpuk.atlassian.net/browse/KBR-2).

**Two documents, no code.** A test-suite design and its implementation plan, plus one line removed from `.gitignore` so `.system_design/` can be tracked at all. No test code, no product code, no behaviour change.

Kitty Bridge sells three promises. Two of them are the reason anyone chooses it over pointing an agent straight at a provider, and until now **neither was written down anywhere as something we check**:

1. **We don't touch your conversation.** What Claude Code sends is what the provider gets — except where translation genuinely requires otherwise.
2. **The provider can't tell we exist.** A coding plan is priced for a coding agent. If a provider can fingerprint bridge traffic it can throttle or terminate the account, and the user finds out mid-session.
3. **If you configure a shared egress IP, nothing escapes it.** Not one request in a thousand.

### `TEST_SUITE.md` — the design

Three named, checkable invariants and the four-layer suite that proves them, from millisecond unit tests up to sessions with real agent binaries.

- **A 35-row register** of every place the bridge is permitted to alter or redirect an agent's request, each with the condition under which it fires and why it is unavoidable. This is what turns "we don't change your messages unless necessary" into something a test can fail.
- **A fidelity oracle built from independent readers**, not from the code under test. Every request projects into a common form — envelope, conversation, residual — and any field a reader cannot account for **fails the run**, because an unaccounted field is exactly where an unregistered change hides.
- **A containment test that proves the negative**: with the proxy off, the provider must receive *zero* connections. Nothing checks this today, so a bridge that quietly fell back to a direct route would pass our entire current suite.
- **Falsification controls on both harnesses.** A test never shown to fail is indistinguishable from one that *cannot* fail.
- 19 gaps, priority-ordered. 14 open questions I did not invent answers to.

### `TEST_SUITE_IMPLEMENTATION_PLAN.md` — the plan

**98 atomic tasks** across a shared-contracts milestone and ten epics, arranged so parallel work starts early.

- **Milestone 0 is contracts plus one proven vertical slice** — types, register schema, corpus format, recorder and fixture extension interfaces, exemption registry, CI selection rules, and one request driven end to end so the contracts are known to compose before six streams build on them.
- **The schedule is derived, not asserted.** The epic tables are the only place dependencies live; tiers and chains are computed from them. 98 tasks, acyclic, every dependency resolves. 43 tasks are available in the first three tiers.
- **Durations are dependency-only lower bounds** — they assume unlimited parallel staffing, which the plan's own constraints break. Stated as such rather than presented as a forecast.

## Five defects found while writing the design

We hadn't written a single test. Specifying carefully enough to be testable was sufficient to surface these — three of which break the promises above in the shipping product today:

| | What | Impact |
|---|---|---|
| [KBR-5](https://shelpuk.atlassian.net/browse/KBR-5) | The literal string **"Kitty Bridge"** is sent to the provider inside the request body | Breaks promise 2 outright |
| [KBR-6](https://shelpuk.atlassian.net/browse/KBR-6) | Two internal fields reach the provider on **every** Chat-Completions provider | Breaks promises 1 and 2; confirmed against six providers |
| [KBR-7](https://shelpuk.atlassian.net/browse/KBR-7) | OpenCode Go declares one wire format and sends another for most models | Latent bug in the thinking-repair path |
| [KBR-8](https://shelpuk.atlassian.net/browse/KBR-8) | The ChatGPT-subscription path claims **two different client versions in one request**, one of them our own release number | The sharpest fingerprint in the codebase |
| [KBR-9](https://shelpuk.atlassian.net/browse/KBR-9) | The README documents a Gemini endpoint that doesn't exist | Anyone following it gets a 404 |

None is fixed here. KBR-5 and KBR-6 are small and ranked priority 0.

## Review history

Four rounds on the design, two on the plan, 37 findings, all adopted, none rejected.

The recurring pattern is worth stating, because it is the reason the documents look the way they do: **nearly every finding was a test that would have passed without proving anything.** A fidelity check that could not detect a changed model name, in a product whose entire function is changing the model name. A containment test asserting nothing arrived at a destination that could not be reached at all. A guard confirming a function was *called*, when that function only returns advice and the enforcement is the three lines after it. A failure report that could not run in the one case it existed for. Several of those were mine, introduced while fixing the previous one.

That is why every harness now ships with a deliberate-breakage case, and why the plan's schedule is computed rather than written — the same failure mode, applied to a document instead of a test.

## Note on `.gitignore`

`.system_design/` was ignored, so the file KBR-2 asks for could not be committed or reviewed. I removed that one line and left `.requirements/` ignored. Flagged as Q8 in the design — your call, easily reverted.

## Testing

Documentation only. `ruff check .` passes; the full suite is unaffected. A cross-reference pass confirms every identifier in both documents resolves and no section reference dangles. The plan's dependency graph is verified acyclic with no undefined predecessors.

https://claude.ai/code/session_01Kq96s1XUMZtR9fxtoQJNeQ

[KBR-2]: https://shelpuk.atlassian.net/browse/KBR-2?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[KBR-5]: https://shelpuk.atlassian.net/browse/KBR-5?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,6 @@
 /.import_linter_cache/
 /.requirements/
 /.serena/
-/.system_design/
 /.venv/
 /debug/
 /.codex
--- a/.system_design/TEST_SUITE.md
+++ b/.system_design/TEST_SUITE.md
@@ -186,86 +186,176 @@ trigger no test can fail — so each material mutation gets its own row.
 | P9a | Set `User-Agent` to `claude-code/1.0` | `KimiCodeAdapter`, `BytePlusAdapter`, `MimoAdapter` `.build_upstream_headers` | Always, on those three | Those providers 403 without a recognised coding-agent user-agent. Central to I2 — F1. |
 | P9b | Remove `Authorization`, add `api-key` | `MimoAdapter.build_upstream_headers` | Always | MiMo does not use Bearer auth. An auth-**scheme** change §4.3 C1's exact-set assertion must encode. |
 | P9c | Synthesise a Codex CLI `User-Agent` and a `version` header | `OpenAISubscriptionAdapter` | Always | Impersonation required by the subscription endpoint. **The two disagree — see F1.** |
-| P10 | Set `reasoning_split = True` | `MiniMaxAdapter.normalize_request` | **Unconditionally** | Makes MiniMax return thinking in `reasoning_details` instead of inline tags. Unconditional, so exempt from §3.3 assertion 2. |
-| P11 | Translate CC → Bedrock Converse | `BedrockAdapter.translate_to_upstream` | Always, on `bedrock` | A third upstream wire format M2 does not name. Custom transport — see §3.3.2. |
+| P10 | Set `reasoning_split = True` | `MiniMaxAdapter.normalize_request` | **Unconditionally** | Makes MiniMax return thinking in `reasoning_details` instead of inline tags. Unconditional, so exempt from §3.3.2 assertion 2. |
+| P11 | Translate CC → Bedrock Converse | `BedrockAdapter.translate_to_upstream` | Always, on `bedrock` | A third upstream wire format M2 does not name. Custom transport — see §3.3.4. |
 | P12 | Translate CC → Ollama `/api/chat` | `OllamaCloudAdapter.translate_to_upstream` | Always, on `ollama_cloud` | A fourth wire format. Custom transport. |
+| P13 | **Drop fourteen Chat-Completions-only parameters** — `temperature`, `top_p`, `max_tokens`, `max_completion_tokens`, `frequency_penalty`, `presence_penalty`, `logprobs`, `top_logprobs`, `response_format`, `stop`, `n`, `stream_options`, `seed`, `logit_bias` | `OpenAISubscriptionAdapter._cc_to_responses` | Always, on the **CC-origin** path | The Codex backend applies strict allowlist validation and rejects them with 400. **User-visible**: `max_tokens` and `temperature` silently do nothing on this provider. Logged at DEBUG. |
+| P14 | **Drop every parameter outside the Codex allowlist**, notably `max_output_tokens` | `OpenAISubscriptionAdapter._prepare_responses_body` | Always, on the **Responses-origin** path | Same backend restriction, different input shape — this path receives a Responses body, so the parameter is spelled `max_output_tokens`, not `max_tokens`. A single row cannot cover both paths; the sets differ. |
+| P15 | **Strip `strict` from every tool declaration** | `_prepare_responses_body` | Always, on the Responses-origin path | The Codex backend rejects it. A change to the **tool schema** the agent declared, not to a sampling parameter — a different kind of fidelity mutation and worth its own row. |
+| P16 | Rewrite content types `input_text` → `output_text` | `_convert_content_types`, called from `_prepare_responses_body` | Always, on the Responses-origin path | The Codex backend validates content types strictly. **Message-content mutation.** |
+| P17 | Inject `stream: True` and `store: False` | `_cc_to_responses` and the Responses-origin body builder | **Unconditionally**, both subscription paths | The Codex backend is streaming-only; kitty reassembles a non-streaming reply from the SSE. Note `stream: True` **overrides a non-streaming client request** — the subscription-path analogue of M11. |
+| P18 | Remove `modelId` and `stream` from the Converse payload | ``BedrockAdapter` transport (`make_request` / `stream_request`)` | Always, on `bedrock` | boto3 takes the model id as a call argument and selects streaming by choosing `converse` vs `converse_stream`, so both must leave the body. Applied **in the transport, after `translate_to_upstream`**. |
+| P19 | Overwrite `stream` | `OllamaCloudAdapter` transport (`make_request` sets `False`, `stream_request` sets `True`) | Always, on `ollama_cloud` | The transport, not the caller, decides which Ollama endpoint mode is used. Applied **after `translate_to_upstream`** has already set it from the request. |
 
 **Conditional rows are the point.** Every row whose trigger is a condition must be provably
 *inert* when that condition is absent — the sharpest form of "unless absolutely necessary", and
-what §3.3 assertion 2 tests. M1, M2, M10, P1, P6, P9a–c, P10, P11 and P12 are unconditional by
-design and are exempt from that assertion.
+what §3.3.2 assertion 2 tests, with the trigger complements §3.3.4 requires. M1, M2, M10, P1,
+P6, P9a–c, P10, P11, P12, P13, P14, P15, P16, P17, P18 and P19 are unconditional by design and
+are exempt from that assertion.
 
 **Register maintenance.** The register is the specification. A pull request that adds a mutation
 site without adding a row fails the L2 register guards (§6.2.3).
+
+#### 3.2.3 Serialization paths — where the register must actually be checked
+
+The register is only enforceable at the point where bytes are handed to a transport. That point
+is **not** `translate_to_upstream` for any of the three custom-transport providers, and assuming
+it is hid seven rows in the first draft: on `openai_subscription` the hook returns a Chat
+Completions shape and the real body is built afterwards (P13–P17), while on `bedrock` and
+`ollama_cloud` the hook builds the body and the transport then **mutates it** (P18, P19).
+
+| Path | Adapters | Where the final bytes are decided |
+|---|---|---|
+| Bridge aiohttp | the 20 default-transport adapters | The `json=` body passed to `session.post` in `_make_upstream_request` / `_open_upstream_stream` |
+| curl_cffi | `openai_subscription` | Built **inside the transport**: `_cc_to_responses` (CC-origin) or `_prepare_responses_body` (Responses-origin). The adapter hook's output is not what ships. |
+| botocore | `bedrock` | Built by `translate_to_upstream`, then **mutated** in `make_request` / `stream_request` (P18). Capture after the mutation. |
+| provider aiohttp | `ollama_cloud` | Built by `translate_to_upstream`, then **mutated** in `make_request` / `stream_request` (P19). Capture after the mutation. |
+
+Every guard and every oracle run in this document targets the right-hand column, never the hook
+that precedes it. Where the body is built in the hook and mutated in the transport, "after the
+mutation" is the boundary — capturing the hook's return value would miss P18 and P19 exactly as
+it missed P13.
 
 ### 3.3 The transparency oracle
 
 The single piece of new infrastructure that makes I1 testable.
 
 **What it is.** A test harness that runs a request through the real `BridgeServer` into a
-recording fake upstream, then diffs the captured upstream body against the inbound agent body
-and classifies every difference against the register.
+recording upstream, projects both the inbound agent body and the captured upstream body into a
+**wire-independent semantic form**, and classifies every difference against the register.
 
 ```
-inbound agent body ──► BridgeServer ──► recording upstream
-        │                                      │
-        └───────── structural diff ────────────┘
-                          │
-                classify each delta
-                          │
-     ┌────────────────────┴────────────────────┐
-claimed by a register row              unclaimed  ──► FAIL
-whose trigger was present                    │
-            │                          unclaimed delta is the
-    assert the row's                   whole point of the oracle
-    stated shape held
+inbound agent body ─────► BridgeServer ─────► recording upstream (final wire bytes)
+        │                                                 │
+   project()                                         project()
+        │                                                 │
+        ▼                                                 ▼
+  Conversation                                      Conversation
+        └──────────────── structural diff ────────────────┘
+                              │
+                    classify each delta
+                              │
+         ┌────────────────────┴────────────────────┐
+   claimed by a register row              unclaimed  ──► FAIL
+   whose trigger was present                    │
+                │                          unclaimed delta is the
+        assert the row's                   whole point of the oracle
+        stated shape held
 ```
 
-**The two assertions, in priority order.**
+#### 3.3.1 The comparison is an independent projection, never a round-trip
 
-1. **No unclaimed delta.** Any difference between inbound and upstream bodies must map to a
-   register row. This is the invariant.
-2. **No mutation without its trigger.** For each conditional row, a corpus entry that does not
-   meet the trigger must show that row's mutation *absent*. This is what stops M3/M5 quietly
-   becoming unconditional.
+An earlier draft of this design specified comparing the inbound body against a **round-trip**
+through the production translators — `translate_request` then `translate_response`. That is
+wrong twice over, and the reason is worth recording so nobody proposes it again.
 
-#### 3.3.1 The oracle is scoped by wire shape — and why
+**It is a category error.** `MessagesTranslator.translate_request` maps a Messages *request* to a
+Chat Completions *request*. `MessagesTranslator.translate_response` maps a Chat Completions
+*response* — a body whose payload is `choices` — back to a Messages *response*. They are not
+inverses and never were: composing them takes a conversation in and returns an assistant reply,
+so the input history cannot survive the trip. The same objection applies to any "inverse pair"
+assembled from the provider adapters.
 
-A direct inbound-vs-upstream diff is only meaningful where the two bodies are in the same
-protocol. On a Chat-Completions-wire provider the upstream body is CC while the inbound body is
-Anthropic Messages: *every* field differs, every delta is claimed by M2, and assertion 1 passes
-trivially while proving nothing. A test that cannot fail is worse than no test.
+**Even a real inverse would prove the wrong thing.** Using the production translator to check the
+production translator establishes self-consistency, not fidelity. A translator that drops the
+same field in both directions round-trips perfectly. The oracle must not be written in terms of
+the code under test.
 
-| Path | Adapters | Comparison | Why |
-|---|---|---|---|
-| **Native Messages wire** | `anthropic`, `zai_coding`, `custom_anthropic`, `minimax_token`, and `opencode_go` **for its Messages models only** | Inbound body vs. upstream body, directly | Same protocol both sides. "Unchanged" is literally meaningful, and this is the path the Jira premise names. |
-| **Chat Completions wire** | the remaining adapters | Inbound vs. **round-trip** (`translate_request` → `translate_response` back to the agent's protocol) | The claim that survives translation is semantic preservation, not byte equality. |
-| **Other wire** | `bedrock` (Converse), `ollama_cloud` (`/api/chat`) | Round-trip through that adapter's own pair | Neither Messages nor CC. Round-trip mode needs the adapter's own inverse, not the shared one (P11, P12). |
+**The projection.** Define one **`Conversation`** value type in the test tree:
 
-**The scoping predicate is per-request, not per-adapter.** `upstream_wire_is_messages_api` is the
-natural selector, but it cannot be trusted as an adapter-level constant: `OpenCodeGoAdapter`
-inherits `True` from `AnthropicAdapter` while its `translate_to_upstream` returns a **Chat
-Completions** body for every model outside `_MESSAGES_MODELS`. See finding F5. Until that is
-fixed, the oracle must select on the observed body shape, and an L2 guard must assert the
-property agrees with the actual output shape for every adapter × representative model.
+```
+Conversation
+  system:    ordered text parts
+  turns:     ordered [ Turn(role, parts) ]
+  tools:     ordered [ ToolDecl(name, schema) ]
+  sampling:  { max_tokens, temperature, top_p, stop, … }   # declared, may be absent
 
-#### 3.3.2 The oracle must be parametrised over transport
+Part = Text(str)
+     | ToolUse(id, name, arguments)
+     | ToolResult(tool_use_id, content, is_error)
+     | Thinking(text)
+     | Image(digest)
+```
 
-`openai_subscription`, `bedrock` and `ollama_cloud` set `use_custom_transport = True` and never
-go through `_make_upstream_request`, so §7.2's aiohttp recording upstream never sees their
-bodies. The oracle is parametrised over transport exactly as §5.5's containment harness is:
-an aiohttp recording server for the default path, a `curl_cffi`-reachable one for the
-subscription provider, and a botocore endpoint override for Bedrock. Without this, §3.4's "every
-provider" rows are false for three adapters.
+Then write one **hand-written projection per wire format** — Anthropic Messages, Chat
+Completions, OpenAI Responses, Gemini, Bedrock Converse, Ollama `/api/chat` — each written
+directly against that format's published shape and **importing nothing from `src/kitty/bridge`**.
+Six small readers, each independent of whatever kitty code produces that format. The oracle compares
+`project(inbound)` with `project(captured_upstream_bytes)`.
 
-**Diffing semantics.** Structural, not byte-level. JSON key order and whitespace are not part of
-the agent's meaning and vary with serialisation; a byte diff would fail constantly and teach
-people to ignore it. The oracle compares parsed JSON with a canonical ordering, and reports
-deltas as JSON-pointer paths so a failure names the exact field. The one exception is key
-ordering on the native path, which §4.3 C2 asserts for I2 reasons.
+This is the independent-oracle rule, and it is what makes the check meaningful across a protocol
+boundary: a Messages body and a Chat Completions body are not comparable as JSON, but their
+projections are directly comparable, because a conversation is a conversation.
 
-**What it does not do.** It does not judge whether a translation is *semantically* correct — that
-a `tool_use` block became the right `tool_calls` entry is an L1 translator claim. The oracle
+**Response translation is tested separately**, with its own projection (`Reply(parts, stop_reason,
+usage)`) over the response direction. It is a different claim and it gets a different test.
+
+#### 3.3.2 The two assertions
+
+1. **No unclaimed delta.** Every difference between the two projections must map to a register
+   row whose trigger the input met. This is the invariant.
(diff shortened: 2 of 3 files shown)


shelpuk:
## Review adopted in full — pushed in 9302198

Thank you, this was a substantially better document afterwards. All twelve findings are adopted. I verified each against the source before acting on it rather than taking any on trust, and three I reproduced directly. Nothing was rejected on the merits.

Four of the twelve were the serious kind — **tests that would have passed without proving their claim**, which is worse than no test because it manufactures confidence:

| Finding | Why it could not have worked |
|---|---|
| The fidelity oracle | `translate_request` maps request→request; `translate_response` maps response→response. Not inverses. The round-trip could never have recovered the input conversation — and even a real inverse would only prove the translator agrees with itself. Replaced with hand-written projections per wire format into a common `Conversation` value, importing nothing from `kitty.bridge`. |
| The egress negative test | Asserted zero upstream connections against an unresolvable `.invalid` host. A broken implementation attempting a direct fallback fails DNS and satisfies it. This is the same class of trap as the `127.0.0.1` one already documented — and it was created by the fix for that one, which is worth recording. Now bracketed by a positive control (prove direct reachability first, per transport) and a falsification control (inject a deliberate bypass, prove the harness fails). |
| The register guard | Inspected `normalize_request` + `translate_to_upstream`. All three custom transports change the body **after** those hooks. Moved to the serialization boundary, per transport. |
| Streaming recovery | Checked SSE grammar only. Duplicated text, replayed tool calls and tool arguments spliced across two attempts are all grammatically valid and all corrupt the conversation. Four failure-injection points with content assertions. |

And two would have **failed on correct behaviour**, which is how gates get disabled:

- The blanket `kitty` body scan would reject a user asking Claude Code to explain kitty-bridge — and satisfying it by stripping their words would breach message fidelity. The two invariants were in direct conflict and I had not noticed. Now scoped to bridge-introduced content via the projection diff, with a regression case asserting both halves in one run.
- `password not in repr+str+masked` false-fails when the password equals a substring of the host. Reproduced with password `proxy` and host `proxy.example`. Now structural.

### Your point 3 was the most productive

Checking the serialization boundary rather than the adapter hook surfaced **seven** unregistered mutations, not one. You were right that `_cc_to_responses` drops parameters — and it drops **fourteen**, not thirteen; I had missed `top_logprobs`. Beyond that:

- `_prepare_responses_body` drops a *different* set on the Responses-origin path (`max_output_tokens`, not `max_tokens`), so one row could not cover both sites.
- It also strips `strict` from every tool declaration — a change to the agent's tool schema, not a sampling parameter.
- It rewrites content types `input_text` → `output_text`.
- Both subscription paths force `stream: True` unconditionally, overriding a non-streaming client request.
- Bedrock's transport pops `modelId` and `stream` *after* `translate_to_upstream` builds the payload.
- Ollama Cloud's transport overwrites `stream` the same way.

The last two also corrected my §3.2.3 table, which described those two bodies as built inside the transport when they are built in the hook and then mutated. Now P13–P19, and the boundary is stated as "after the mutation."

### Deferred — decisions and numbers only, each with an owner

I adopted every design correction. What I did **not** do is invent answers to things that are genuinely the product owner's call or that need a measurement. These became explicit open questions rather than quiet assumptions:

| Deferred | Why | Where |
|---|---|---|
| What *should* happen when the final turn alone exceeds the budget | Your point 6 exposed undesigned behaviour, not a wording slip. Writing it up as intended would have ratified an accident. The properties and TR-3 now record current behaviour as **observed, explicitly not approved**, and move together when this is answered. | **Q10** |
| Whether changed-code mutation testing fits the per-PR gate | You are right that rejecting it needs a measured runtime, not an assumption. The nightly-only choice is marked provisional pending that number — measured against a gate that already runs ~18.5 min × 4 Python versions. | **Q11** |
| How a pinned Claude Code binary reaches CI | I adopted the requirement — an arbitrary child process cannot test Claude Code's own settings precedence, which is exactly the assumption under suspicion in KBR-1. But distribution, pinning and redistribution licensing are real questions. If the answer is "not feasible", that limitation gets stated rather than papered over. | **Q12** |
| Baseline for the compaction arm of the evals | The direct-provider arm returns a 400 on an over-context input, so there is nothing to compare. Candidates are a larger-context model or relaxed compaction; the choice changes what a regression there means. | **Q13** |
| Load ceilings and the eval regression margin | These come from a first recorded baseline, then ratchet. Inventing them would be worse than leaving them open — and a pre-registered margin has to be chosen before the data, not after. | **Q4**, §6.4.4 |

### On the ratchet

Your point 8 surfaced something I had left inconsistent: TR-1's header-subset clause and TR-4 assert the *target* state of two currently-open defects (KBR-8 and KBR-5), and I had placed them in a job that gates every PR. They would have made `main` red on day one, and a permanently red gate gets switched off, taking the working scenarios with it. Both are now `@ratchet` — they run and report, they do not gate, and the marker comes off when the defect closes.

### Verification

Every register row and serialization path re-checked against `openai_subscription.py`, `bedrock.py` and `ollama_cloud.py`. A cross-reference pass confirms all M1–M13, P1–P19, F1–F5, G1–G18 and Q1–Q13 identifiers resolve, and no `§` reference points at a missing section. `ruff` passes; the change is documentation only.

Two follow-on findings from this round were unregistered-mutation issues rather than defects — necessary behaviour that was simply invisible above DEBUG — so they are register rows plus **Q5** on user visibility, not new tickets. I did not want to file noise.

shelpuk:
## Second review adopted in full — pushed in 945060c

All seven adopted, plus your closing point on streaming semantics. Nothing rejected. You also correctly flagged that "adopted all twelve" and "resolved all twelve" were different claims — that was fair, and finding 1 is the proof.

### Finding 1 was the serious one, and it landed on my own fix

The projection I introduced last round to replace the broken round-trip kept only semantic content. It had no `model` field. So it could not see **M1** — the model override, which is the entire reason this product exists — nor P6, P15, P17, P18 or P19. Both sides would project identically and "no unclaimed delta" would pass while the mutation went upstream. The vendor-token check inherited the same blind spot: bridge-added metadata discarded during projection could never have been scanned for a vendor string.

I had replaced a test that could not run with a test that could not fail on the most important case. That is a worse failure than the one I was fixing, and it is exactly the class of error this document spends its length warning about.

The projection is now total:

```
Request
  envelope:     model, stream, store, and every control field the format defines
  conversation: system, turns, tools(name, description, schema, strict), sampling
  residual:     anything the reader did not account for
```

Every key in a body must be classified into exactly one. **A non-empty residual fails the run** — not reported as a diff, not ignored. Adding a field to a wire format now forces a decision: map it, or declare it ignored with a reason. Every register row names the field it touches, so "claimed by a row" is a lookup, not a judgement.

And the oracle now carries a falsification control of its own, matching the one §5.2.2 already demands of the containment harness: five injected mutations — changed model, flipped `stream`, deleted tool description, stripped `strict`, injected metadata field — must each go red, in the suite, not once by hand. I should have applied that symmetry the first time; I had required it of the egress harness and not of the fidelity one.

### The rest

| | Adopted as |
|---|---|
| **2** Eval denominators | Successes ÷ **scheduled** trials. Your 90-refusals example is decisive: excluding them reports a catastrophic regression as a clean run. Refusals are evidence *about the arm* — changed instructions and lost context both cause them, and both are precisely what compaction does. Failure taxonomy reported separately, exclusions only by a pre-written symmetric rule, and a missing-data ceiling that **voids** the run rather than adjusting it. |
| **3** Agent smoke | You're right that connectivity is not precedence — with nothing competing, an implementation with the order backwards passes. Now sets conflicting values in the process environment and in `~/.claude/settings.json`, each pointing at its own sentinel recorder, and asserts both stay **silent**. Plus a control run proving the sentinels can be hit, so a typo in a sentinel URL doesn't read as a pass. |
| **4** Mutation scope | Added `_cc_to_responses`, `_prepare_responses_body`, `_convert_content_types`. P18/P19 sit inside network methods and can't be reached from an L1 selection, so the payload shaping gets extracted into pure builders first — the factoring is what blocks the test, not the test being hard to write. |
| **5** Release wiring | Correct, and it would have shipped: `publish.yml` depends only on `tests.yml`, so a separate load workflow leaves publication free to proceed while load fails, or while it never ran for that commit. `load.yml` becomes reusable and `publish.yml` calls it, gated on the tag commit. A row in a table is not a dependency. |
| **6** `@ratchet` | It was a blanket amnesty — a broken fixture or an unrelated regression in those scenarios would have been as invisible as the known defect. Now exempts **one named assertion** with its issue key; setup and every other assertion gate normally; an unexpected pass **fails** the job so the exemption cannot outlive its defect. |
| **7** Rationale and PR body | §10 still claimed pairing cancels model variance while §6.4.3 said the opposite. Fixed: pairing holds task difficulty constant, it does not cancel sampling. PR description updated — the register is 39 rows, the gap list 18, the questions 14. |

### On the streaming point

You were right not to let this pass. "Whichever the agreed semantics say" left that row without an acceptance oracle, which is the same defect the document objects to elsewhere. It is now **Q14**, and the table says so explicitly: until it's decided, that row asserts only the negatives — no duplicated text, no reused tool-call id across attempts, no spliced arguments. That catches corruption but cannot confirm correctness, and the design now admits it rather than implying the check is complete.

I've also marked Q10–Q14 as implementation prerequisites in the question list, so they stay visible as blockers rather than reading as general musings.

### Standing

Adopted, verified, pushed. Every code claim was checked against the source before acting — including finding 1, where I confirmed the missing `model` field would have hidden M1. Cross-reference pass is clean: all M1–M13, P1–P19, F1–F5, G1–G18 and Q1–Q14 resolve, no dangling `§` references.

shelpuk:
## Third review adopted — pushed in e377d0e

Both P2 items addressed, both P3 corrections applied. Nothing deferred from this round.

### 1 — Routing was missing from the register, not just from the oracle

You framed this as an oracle-interface gap. Checking it against the code, it was wider: the destination is built from the profile by `build_base_url()` + `get_upstream_path()` and **no register row covered it at all**. So the design had no claim about where a request goes, only about what it carries.

Azure is the case that makes it concrete, and it is sharper than I had appreciated. P6 removes `model` from the body *because* the deployment id lives in the URL. So two requests to two different Azure deployments — different models, potentially different billing — have **byte-identical bodies**. A body-only oracle cannot distinguish them by construction. The register also read as if Azure *dropped* the model, when it *moved* it, and the move was unchecked on both sides.

Added:

- **M14** — the destination is replaced wholesale; the agent addressed a loopback bridge and the request has to reach a real provider.
- **P20** — Azure encodes the profile's model as the deployment id in the path. Explicitly the counterpart of P6, so the pair reads as a move rather than a deletion.
- **P21** — Vertex encodes `project_id` and `location` in the base URL.
- **§3.3.5** — the oracle now takes complete captured requests (method, scheme, host, path, query, headers, body).

On your point about deriving expectations independently: agreed, and it is the same rule §3.3.1 already applies to bodies. The expected route is computed in the test from the profile using the provider's published URL shape — never by calling `build_base_url()` / `get_upstream_path()`. Asking the code under test where it intended to go and then confirming it went there is not a test. Sixth falsification case added: change the deployment segment, leave the body byte-identical, the oracle must fail.

You were right that the recorders already capture most of this. §7.2 now lists `query` explicitly and says why the routing fields are load-bearing rather than incidental — that omission is presumably how the assertion came to ignore them.

### 2 — Calling the guard is not enforcing it

Correct, and this one would have survived every check in the document. `egress_block_reason()` returns a reason; it stops nothing. The enforcement is the branch after it:

```python
egress_error = egress_block_reason(provider, profile, resolved_key, backends)
if egress_error:
    print(f"Error: {egress_error}", file=sys.stderr)
    return 1
```

Delete those three lines and AST domination still passes, the guard's own unit test still passes, and an unproxyable profile starts and leaks the machine's address — precisely the failure I3 exists to prevent. A structural check can prove a call site exists; it cannot prove the answer is obeyed.

Paired it with a behavioural check at L3: each of the five start paths driven with a rejecting configuration, asserting **no server starts** — no listening socket, non-zero exit — not merely that a message was printed. Plus a falsification control that keeps the call and discards its return value, which must make those tests fail. The three now divide cleanly: the unit test proves the guard *decides* correctly, this proves the decision is *obeyed*, EG-3 proves the user sees something useful.

### 3 — The control exercised one sentinel, not two

Right, and the failure mode is quiet: A is *supposed* to be silent in the main run, and a broken A is silent too, so a typo in its URL reads as a pass. Control 1 only ever demonstrated B. Third run added — session settings and the global override both absent, so A must receive the request. Three runs, three winners, every destination shown live under its own configuration.

### 4 — Stale acceptance policy

Caught. §8 had moved to assertion-level exemptions but §6.4.1 still said the scenarios "do not gate until their defect closes", and the Gherkin comments carried the same reading. Implementing §6.4.1 literally would have rebuilt the blanket amnesty you flagged last round. §6.4.1 now defers to §8's registry rather than restating the policy — one statement of it, so the two cannot drift apart again. The scenario-level `@ratchet` tags are gone; the comments name the single exempt assertion and its ticket.

### Standing

Cross-reference pass clean: M1–M14, P1–P21, F1–F5, G1–G19, Q1–Q14 all resolve, no dangling `§`. New gap **G19** records the routing hole at priority 1, alongside the two invariants with the least coverage.

Noted and agreed on your closing caveat: this is a static design review, and none of it establishes that the proposed harnesses work. The falsification controls exist precisely because that only gets settled when someone builds them and watches them fail.

shelpuk:
## Added: the implementation plan — `b81ba47`

`.system_design/TEST_SUITE_IMPLEMENTATION_PLAN.md`. 84 atomic tasks implementing the design, arranged so parallel work can start early. Still design only — no test code, no product code, and no Jira issues yet.

### Organised around what unblocks the most people soonest

The design has an awkward shape for delivery: almost every interesting test depends on shared infrastructure that does not exist. Build it in the wrong order and one person writes harnesses for three weeks while everyone else waits.

**Wave 0 is eight small artifacts**, chosen purely for how much each unblocks:

| | Unblocks |
|---|---|
| Layer markers | Everything — nothing lands in the right CI job without them |
| Request / projection types | All six format readers, then the oracle |
| **The register as data** | It is a markdown table today, and both the oracle and the completeness guard consume it — so it becomes data or both parse prose |
| Recorder protocol + primary recorder | The other three recorders, the bridge fixture, all of L3 |
| Shared CONNECT proxy | The entire containment stream |
| Corpus capture + scrubber | All corpus entries |
| **Exemption registry** | Every guard designed to land red |
| **Bridge-under-test fixture** | ~15 tasks across five epics |

Parallel capacity: ~10 during Wave 0, **~15 in Wave 1**.

**Two long chains, and they are independent.** Fidelity (~15 days) and containment (~17–18 days) converge only at the CI wiring, so two owners running them in parallel roughly halves elapsed time on the two hardest parts of the suite. That is the largest scheduling lever in the plan.

### What review caught

Every one of these produces a task someone could not have completed cold:

- **Identifier collision.** My bare task ids reused the design's namespaces — `C1–C6` are observable channels there, `F1–F5` findings, `G1–G19` gaps, `I1–I3` the invariants. The first draft said "KBR-6 → G3" where the design says "KBR-6 → **G15**". Everything is now `T-` prefixed with a Design column carrying the reverse link.
- **Two Wave 0 artifacts were missing entirely.** No task delivered the bridge-under-test fixture that ~15 tasks need — so T-D1, T-E3, T-G2 and T-I8 would each have built their own. And the exemption registry sat in Wave 3 while four guards that cannot land without it sat in Waves 0–2.
- **Four design specifications had no task at all**: the header contract (§4.3 C1 — the test that catches KBR-8), fingerprint parity (C1b), side traffic (C6), and the byte-level key-order assertion (C2).
- **Wrong dependencies in five places.** A1–A6's acceptance required a corpus they did not depend on, which also dragged the corpus onto the critical path; D1 needed scripted responses that W0.4 explicitly excludes; G2 omitted four projections and the whole corpus; J2 carried a containment dependency no TR scenario uses; E1 claimed per-transport routing while depending only on the aiohttp recorder.
- **The critical path was wrong twice.** The containment chain is longer than the fidelity chain, and neither ends at the acceptance scenario — a scenario that gates nothing is not delivered, so both chains end at the CI wiring.
- **One task would have broken CI.** T-K6's Acceptance selection includes `agent_smoke`; if the Q12-blocked task never lands, the selection collects nothing, pytest exits 5, and the skip-is-failure rule turns that into a hard failure on every PR.
- Smaller: H2's "behaviour unchanged" now depends on the characterisation test that makes those bytes observable; W0.1 defaults markers by path rather than hand-editing 2,880 tests across 139 files — the worst possible first commit when a dozen people are about to branch; corpus refresh has a named owner.

### Two things carried over deliberately

**Guards land red, on purpose.** Four guards and two scenarios assert the target state of KBR-5 … KBR-9. Writing a guard *after* its fix proves nothing about whether it would have caught the defect — and the fix then has no regression test that ever failed. Each lands with a single-assertion exemption; the defect's own ticket removes it, and the unexpected-pass rule makes forgetting impossible.

**Every harness needs a committed falsification case.** Not "the author verified it once" — a deliberate defect the harness must detect, running in the suite. Four review rounds on the design produced four harnesses that would have passed without proving anything; this is the rule that stops the fifth.

### Still blocked on decisions

Q12 (pinned Claude Code binary in CI) is now the most expensive open question — it blocks two tasks *and* reaches into the CI wiring through T-K6. Q4, Q10, Q11, Q13 and Q14 each block or weaken one task; §15 lists what each costs while it stays open.

shelpuk:
## Plan restructured — `10e745e`

All eight findings adopted. The framing was the useful part: I had written a task list and called it a schedule, and those are different artifacts. A schedule has to be *derivable* from the dependencies, and mine wasn't — the graph, the wave table and the duration estimate were three documents that happened to sit next to each other.

### The schedule is now computed

The epic tables are the only place dependencies live; tiers and chains are derived from them. 98 tasks, acyclic, every dependency resolves.

Running that derivation contradicted the previous draft immediately:

> **The critical path runs through the corpus, not containment.**

`T-W2 → T-W3 → T-W6 → T-C1` is four sequential tasks before the oracle can start, because T-D1 needs a corpus entry to run against. I had filed the corpus as a background risk in §18 while it was actually the front of the longest chain. The plan now says to staff it first, and offers an explicit lever — let T-D1 run against a synthetic transcript instead of T-C1 — to take ~2 days off the front if that trade is worth making. That is exactly the kind of decision that should be deliberate rather than accidental.

| Milestone | Days |
|---|---|
| Containment slice proven (T-E2) | 11.5 |
| Fidelity matrix complete (T-D9) | 15.0 |
| Containment acceptance (T-J3) | 16.0 |
| **Everything gating (T-K9)** | **17.0** |

### Your P1s

**1 — T-W8 promised what later tasks supply.** Correct, and the consequence was worse than the wording: completing it literally meant building T-B1–T-B3 *inside* the shared fixture. Split into a core with a **transport extension interface** in Milestone 0, and per-transport integrations that now ship with their own recorder tasks. Recorder authors integrate without editing the core, which is the point you were making.

**2 — The transport streams were serialised for no reason.** T-D5–T-D7 waited on the entire 20-adapter matrix; T-E3 waited on every transport's direct route. As you say, that contradicts this plan's own claim that a stuck transport blocks only its own task. Both are now siblings off a shared interface, with **one complete, falsified aiohttp slice** (T-E2) establishing the pattern before it is replicated.

### Your P2s

**3 — Acceptance not supported by predecessors.** All four confirmed, plus the one you caught in the schedule itself: T-K6 sat on the claimed fidelity chain without depending on T-J2. That is the specific failure mode of keeping dependencies in two places, which is why they now live in one.

**4 — The CI claim was wrong.** You are right and I should have checked it. `pytest -m "acceptance or agent_smoke"` does **not** exit 5 once acceptance tests exist — it collects them and passes with zero agent-smoke coverage. Each required category now has its own non-empty check; the selection mechanism moves to T-W1 where parallel authors need it from day one; and `agent_smoke` is its own category, so Q12 no longer reaches into the Subsystem or Acceptance gates.

**5 — Falsification mandatory in prose, optional in the graph.** T-E4 could have fed acceptance and CI without T-E5, and T-D4's consumers did not depend on T-D3 — so both major harnesses could have become gating infrastructure before their falsification suites existed. Adopted your first option: the **first working version of each harness ships with a falsification case**, and the dedicated suite is a hard prerequisite for that harness gating anything downstream.

**6 — Coverage gaps.** All three confirmed: the response-direction `Reply` projection and comparison (T-A7, T-D10), the expansion from two live-agent cases to the design's five Claude Code scenarios (T-I14 — you are right that scheduling the existing file nightly is not the same as expanding it), and TR-1b. Added §17, a design-to-deliverable coverage table, including the two requirements the **existing** suite already satisfies.

**7 — Merging guards red delays fixes.** This is a correction to a principle I had been defending for several rounds, and you are right. The goal is *evidence that the regression test fails on the unfixed revision* — an atomic PR carrying test and fix, with that evidence in it, satisfies the goal without delaying the fix. Exemptions are now reserved for guards whose scope is broader than one defect, or where the fix must genuinely follow later. What stays prohibited is the thing the rule was actually for: a guard written after the fix with no evidence it would ever have failed.

**8 — Bundled tasks.** Split. The Bedrock and Ollama refactors are now separate, each behind the characterisation capture that makes *its own* bytes observable — bundled, the Ollama half had no evidence at all. Dependency contracts split four ways; nightly workflows split three ways.

### On your closing recommendation

Adopted directly. **Milestone 0 is now "shared contracts plus one proven vertical slice"** — types, register schema, corpus format, recorder and fixture extension interfaces, exemption registry, CI collection rules, and T-W9, which drives one request end to end so the contracts are known to compose before six streams build on them. Shared files get a named integration owner; each stream owns its own modules. Tiers are labelled as a derived view, and §1.1 says explicitly to schedule by readiness rather than waiting for a tier to close.

53 of 98 tasks are available in tiers 0–2.

shelpuk:
## Fifth review adopted — `20f9324`

All six P2 findings closed. Thank you for independently recomputing the graph — that is the check I most wanted and could not perform on my own work.

### 1 — The falsification gate had a second hole

Correct, and instructive: fixing T-J2's dependency on T-D3 felt like closing the finding, but T-K6 activates the gating `l3` job through T-D4 alone, so the oracle could still become gating infrastructure with no falsification suite. **T-K6 now requires T-D3.** One rule, two places it had to be enforced, and I only enforced it in the one you had pointed at.

### 2 — The failure report could not run in the one case it existed for

This was the best finding of the round. T-E9 exists to record that a transport could **not** be proven, and it depended on all three transport tasks completing — where "complete" meant all four phases passing. So a botocore route that could not be established left its task unfinished, and the report designed for exactly that outcome never became ready, taking T-E8 and containment acceptance with it. The mechanism for reporting failure was gated on success.

Restructured as you suggested:

- **T-E1 ships the capability report**, initialised with every transport `not-attempted`.
- **A transport task is done when it records an outcome**, not when it succeeds: `proven`, `unsupported` (the harness cannot route it — with the reason; design §5.3 requires this be reported, not hidden), or `failed` (a route exists and containment does not hold — a product defect, filed).
- **T-E9 depends only on T-E1** and asserts every transport has an outcome with none `not-attempted`. `unsupported` permits partial delivery; `failed` does not.
- **T-E8 moved to the recorders** rather than the containment slices — it observes tunnelling, which needs T-B1–T-B3, not the full slices — so a stuck transport does not block it either.

### 3 — Two foundation tasks owned one contract

Right, and the Gemini detail is what makes it concrete: T-A4 depends only on T-W2 but must read the URL, while `CapturedRequest` was defined in T-W4 with no dependency between the two. Three authors could have started against three readings of one contract — the coordination problem Milestone 0 exists to remove.

`CapturedRequest` now lives in **T-W2**, the input contract. T-W4 implements it and depends on it. T-W9's acceptance asserts the recorder's output is **type-compatible with the declared types**, not merely that a capture happened.

### 4 — Agent-smoke activation was wrong in both directions

Both halves confirmed. T-K10 waited on T-K9 — unrelated fidelity and containment acceptance, no shared prerequisite — while *not* requiring T-I6, so the category could have gone green on connectivity alone without ever proving the settings precedence that is its entire justification. Now `T-W1, T-I5, T-I6`, and the T-K9 dependency is gone.

### 5 — An estimate I asserted instead of computing

You were right that "about two days" was not computed, and the correction is larger than your 0.5 days. Recomputing against the current graph — which now has T-W4 depending on T-W2 from finding 3 — gives **zero days saved**:

```
chain now:    T-W2 -> T-W3 -> T-W6 -> T-C1 -> T-D1 -> ...   17.0 days
without edge: T-W2 -> T-W4 -> T-W8 -> T-W9 -> T-D1 -> ...   17.0 days
```

Dropping the corpus edge makes T-W9 the limiting predecessor at exactly the same length. The corpus dependency is not what makes that path long — the path is long either way. **The lever does not exist and has been withdrawn**, and §18's risk entry now says there is no shortcut rather than pointing at one.

On the framing: adopted. The duration table is labelled a **dependency-only lower bound** assuming unlimited parallel staffing, and **"two owners halve elapsed time" is withdrawn** — that is a resourcing claim the graph cannot support, and the plan itself names two constraints that break it (the shared corpus capture pass, and single ownership of five shared files). What survives is narrower and still true: the two chains share no task until T-K9, so they never block each other.

One consequence worth flagging: containment now carries **~5 days of slack** rather than being nearly level with the critical path. That changes where review attention should go — T-E1 is still the riskiest task, but a slip there is now absorbed rather than fatal.

### 6 — Stale PR description

Fixed. It now describes both documents, the current counts, and the qualified estimates. You are right that readers should not have to reconstruct scope from the discussion history.

### Standing

98 tasks, acyclic, every dependency resolves; tiers and chain durations recomputed and matching the tables. And noted on your closing caveat — this remains a static review of documents. Nothing here establishes that the harnesses work; that is what the falsification cases are for, and they only pay out when someone builds them and watches them go red.

PR #48, opened 2026-09-07
KBR-7: declare the upstream wire shape per model, and guard the whole registry

shelpuk:
Fixes [KBR-7](https://shelpuk.atlassian.net/browse/KBR-7). Closes design gap **G16** / finding **F5**.

Kitty Bridge lets one coding agent talk to any model provider. Each provider adapter declares which API dialect it speaks, and the bridge trusts that declaration when it has to repair a conversation.

**The OpenCode Go adapter was declaring the wrong dialect for most of its models.** It said "Anthropic" but actually speaks "Chat Completions" for the GLM, Kimi and MiMo models. So when a provider rejected a conversation and the bridge tried to repair it, the bridge wrote the repair in the wrong dialect — corrupting the very request it was trying to fix, on the retry path where the user is already having a bad time.

Nothing was silently wrong-but-working: on those models a repair attempt sent a malformed request. This makes the declaration honest per model, and adds a check across all 23 providers so it cannot drift again.

Two adjacent defects were found, filed, and deliberately **not** fixed here — folding them in would have made this three changes wearing one hat:

* **[KBR-126](https://shelpuk.atlassian.net/browse/KBR-126)** — OpenCode Go's routing table is stale against the live provider. Worth a look: the provider now serves **eight** models on its Anthropic endpoint and kitty knows **two**, and there is a whole third endpoint (`/v1/responses`, four models including Grok 4.6 and GPT 5.6 Luna) kitty has no route for. User-visible on six models today.
* **[KBR-127](https://shelpuk.atlassian.net/browse/KBR-127)** — the upstream URL and auth headers are resolved from an un-normalized model name, so a profile written `opencode/minimax-m2.5` sends the right body to the wrong endpoint with the wrong auth.

---

## The defect

`OpenCodeGoAdapter` subclasses `AnthropicAdapter` and inherited `upstream_wire_is_messages_api == True`, while its own `translate_to_upstream` returns a Chat Completions body for every model outside `_MESSAGES_MODELS`.

That property is not decorative. Its own docstring says it "describes the shape that actually goes on the wire" and that anything shaping the serialized body "must branch on this" — and the bridge's thinking round-trip repair does, at two call sites. With `native=True` on a Chat Completions body, `_with_thinking_carrier` rewrites an assistant turn's string content into Anthropic content blocks:

```python
{"role": "assistant", "content": "hi"}
# becomes, inside a Chat Completions body:
{"role": "assistant", "content": [{"type": "thinking", "thinking": ""},
                                  {"type": "text", "text": "hi"}]}
```

The correct Chat Completions carrier is a `reasoning_content: ""` field with `content` untouched. So the repair path was **broken, not merely mislabelled**.

## The fix

`ProviderAdapter` gains `upstream_wire_is_messages_api_for_model(model)`, defaulting to the existing property — the same pairing this codebase already uses for routed adapters (`upstream_path` / `get_upstream_path`, `build_upstream_headers` / `build_upstream_headers_for_model`, both already in this adapter). `OpenCodeGoAdapter` overrides both: the per-model form mirrors its routing predicate exactly, and the bare property reports its default (Chat Completions) route, as the two declarations beside it already do.

Both repair sites read the model from `cc_request`, not `_active_model`. That is what makes declaration and body agree **by construction rather than by coincidence**: `translate_to_upstream` routes on `cc_request["model"]`, and `_normalize_model` normalizes only that key. `.get("model", "")` matches the adapter's own read including its default — a subscript would raise `KeyError` inside the branch that exists to recover from an upstream rejection.

The method is concrete on the base class rather than an optional `hasattr` hook, because the guard must be able to ask *every* adapter; a hook would let it silently skip whatever did not define it.

`AnthropicAdapter`'s property docstring — the sentence that was the origin of F5 — and `ProviderAdapter`'s, which told callers to branch on the bare property, are both corrected.

## Evidence of red

Per `TEST_SUITE_IMPLEMENTATION_PLAN.md` §16, an atomic fix must show its tests failing at the base revision. Reproduced by exporting `origin/main` to a scratch tree and copying only the new test files in:

**At the true base revision** — the bridge tests show the actual corruption:

```
FAILED TestProactiveRepairBranchesPerModel::test_uses_the_chat_completions_carrier_for_a_cc_routed_model
FAILED TestReactiveRepairBranchesPerModel::test_uses_the_chat_completions_carrier_for_a_cc_routed_model
2 failed, 2 passed

E  assert None == ''
E   +  where ... = {'content': [{'thinking': '', 'type': 'thinking'},
                                {'text': 'Reading the file.', 'type': 'text'}],
                    'role': 'assistant'}.get
```

Both Messages-routed tests pass at base, so the pair localises the defect rather than just flagging it.

**With only the base delegation added**, so the sweep can run against the unfixed adapter:

```
FAILED test_declared_wire_shape_matches_the_emitted_body[opencode_go-glm-5.2]
FAILED test_declared_wire_shape_matches_the_emitted_body[opencode_go-kimi-k2.7-code]
FAILED test_declared_wire_shape_matches_the_emitted_body[opencode_go-mimo-v2.5-pro]
FAILED test_declared_wire_shape_matches_the_emitted_body[opencode_go-]
5 failed, 56 passed

E  AssertionError: opencode_go x 'glm-5.2' declares Anthropic Messages but emits chat_completions
```

Red on exactly the four Chat-Completions-routed cases; green for the other 22 providers.

## Tests

Proved at the lowest layer that can prove each claim (`.github/review/rules/python-tests.md`).

**Component** — `tests/test_provider_base.py`, `tests/test_provider_opencode.py`: the base delegation, and that a subclass declaring only the property is still answered correctly (load-bearing for `CustomAnthropicAdapter`); the per-model declaration over both routes; the bare property; and that the declaration is a **predicate, not an observation** — with `translate_to_upstream` monkeypatched to raise, it must still answer, so the registry sweep cannot become a tautology.

**Contract** — `tests/test_wire_shape_honesty.py`, the §6.2.3 guard, in the style of `tests/test_egress_coverage.py`. Sweeps all 23 providers × representative models. Six self-checks so it cannot rot: the classifier pinned against known Messages, Chat Completions and Converse bodies — including a Converse body **without** tools, since the tools axis is what separates Converse from Messages, and a Bedrock Converse body carries a top-level `system` exactly as Messages does; registry coverage; `_MESSAGES_MODELS` coverage; both routes represented for any routing adapter, *and that it has a subject at all*; and the custom-transport set asserted rather than narrated.

**Subsystem** — `tests/bridge/test_thinking_roundtrip_per_model_wire.py`: both repair call sites, each with its Messages-routed complement. The two sites are separate tests deliberately — a fix applied to only one would be **worse than no fix**, because once the bare property reports Chat Completions the site left behind would write `reasoning_content` into an Anthropic transcript. Verified: reverting either site alone reddens exactly its own complement.

The reactive tests use a transcript with no reasoning rather than the sibling module's fixture. This provider is not native-passthrough, so a thinking-enabled request has the carrier injected by `AnthropicAdapter.translate_to_upstream` on the way back through the CC round trip; the repair would be a no-op and the test would fail on a plain 400 instead of on the carrier.

## Scope — what this does *not* deliver

The guard observes `translate_to_upstream`, not the serialization boundary. For the three `use_custom_transport` adapters nothing at the hook observes the bytes that ship: `openai_subscription` never invokes the hook on the request path at all (`_cc_to_responses` builds the Responses body inside the transport), while `bedrock` and `ollama_cloud` mutate the hook's body afterwards (P18, P19) — those two mutations do not change the shape family, so there the exemption is precautionary rather than load-bearing.

The wire-level form remains **T-G4 / KBR-80**. `TEST_SUITE.md` §6.2.3, the T-G4 row, §16 and §17 all now say so explicitly, so a green run here cannot be read as T-G4 delivered.

## Checks

* Full suite: **3144 passed, 5 skipped**, across three independent runs. The only red anywhere is `tests/test_egress_https_proxy.py`, whose fixture needs `openssl x509 -copy_extensions`; reproduced identically on `origin/main`, so it is environmental (Git-for-Windows OpenSSL) and untouched by this diff.
* `ruff check src tests` — clean. `mypy src/kitty` — clean.
* Reviewed by `system-design-reviewer` (three rounds: five blockers and twelve concerns, all resolved) and `code-reviewer` (two rounds, **APPROVE**), each pointed at the requirements and `.system_design/`.

https://claude.ai/code/session_017EWo9i9PUfivaoe8sPL7K8

[KBR-7]: https://shelpuk.atlassian.net/browse/KBR-7?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[KBR-126]: https://shelpuk.atlassian.net/browse/KBR-126?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.system_design/TEST_SUITE.md
+++ b/.system_design/TEST_SUITE.md
@@ -213,9 +213,10 @@ site without adding a row fails the L2 register guards (§6.2.3).
 
 The register is only enforceable at the point where bytes are handed to a transport. That point
 is **not** `translate_to_upstream` for any of the three custom-transport providers, and assuming
-it is hid seven rows in the first draft: on `openai_subscription` the hook returns a Chat
-Completions shape and the real body is built afterwards (P13–P17), while on `bedrock` and
-`ollama_cloud` the hook builds the body and the transport then **mutates it** (P18, P19).
+it is hid seven rows in the first draft: on `openai_subscription` the hook is **never invoked on
+the request path at all** — `_cc_to_responses` builds the Responses body inside the transport
+(P13–P17) — while on `bedrock` and `ollama_cloud` the hook builds the body and the transport then
+**mutates it** (P18, P19).
 
 | Path | Adapters | Where the final bytes are decided |
 |---|---|---|
@@ -375,12 +376,15 @@ not solved the problem.
 
 #### 3.3.4 Scoping, triggers and transports
 
-**Scoped by observed wire shape, per request.** `upstream_wire_is_messages_api` is the natural
-selector but cannot be trusted as an adapter-level constant: `OpenCodeGoAdapter` inherits `True`
+**Scoped by observed wire shape, per request.** The declared wire shape is the natural selector
+but must not be trusted as an adapter-level constant. `OpenCodeGoAdapter` used to inherit `True`
 from `AnthropicAdapter` while emitting Chat Completions for every model outside
-`_MESSAGES_MODELS` (F5, KBR-7). The oracle selects the projection by the shape actually observed
-on the wire, and an L2 guard asserts the property agrees with that shape for every adapter ×
-representative model.
+`_MESSAGES_MODELS` (F5, KBR-7); that is fixed, and the declaration is now per-model. **The
+decision here stands regardless**, for the reason given in §7.4 rather than because of that one
+bug: an oracle must not ask the code under test what shape it emitted, and a declaration is a
+claim, not an observation. So the oracle selects the projection by the shape actually observed on
+the wire, and the L2 guard (§6.2.3) separately asserts that the declaration agrees with that
+shape for every adapter × representative model.
 
 **Triggers and their complements, across representative models.** A single fixed request cannot
 establish register completeness. For every conditional row the corpus must contain a case that
@@ -546,8 +550,10 @@ the test pins it as a *declared* exception and asserts `kitty --no-validate` rem
 
 ### 4.4 Findings
 
-Five defects surfaced while writing this document. **None is fixed by this change**; each needs
-its own ticket. F3, F4 and F5 are live breaches of invariants defined above.
+Five defects surfaced while writing this document. **None was fixed by this change** — it added
+documentation only; each needed its own ticket. F3, F4 and F5 were live breaches of invariants
+defined above. **F5 has since been fixed under KBR-7**, together with the hook-level half of its
+guard; see its entry below and §6.2.3. The rest remain open.
 
 - **F1 — Agent identity is handled per-provider, not by policy.** *(KBR-8.)* Upstream headers are built from
   scratch, so Claude Code's `user-agent`, `x-app`, `anthropic-beta` and `x-stainless-*` never
@@ -586,14 +592,20 @@ its own ticket. F3, F4 and F5 are live breaches of invariants defined above.
   unmistakable proxy signature. Note that `_reasoning_effort` and `_thinking_enabled`, written by
   the same function, *are* in the set — so this is an omission, not a design choice. Tracked as
   G15 and KBR-6.
-- **F5 — `OpenCodeGoAdapter.upstream_wire_is_messages_api` is wrong for most of its models.**
-  *(KBR-7.)* It
-  inherits `True` from `AnthropicAdapter`, but its `translate_to_upstream` returns a Chat
+- **F5 — `OpenCodeGoAdapter.upstream_wire_is_messages_api` was wrong for most of its models.**
+  *(KBR-7 — **fixed**.)* It
+  inherited `True` from `AnthropicAdapter`, while its `translate_to_upstream` returned a Chat
   Completions body for every model outside `_MESSAGES_MODELS`. `ProviderAdapter`'s own docstring
   says the property "describes the shape that actually goes on the wire" and that anything
-  shaping the serialized body must branch on it — so a `True` that is false for most models is a
+  shaping the serialized body must branch on it — so a `True` that was false for most models was a
   latent defect in the thinking-repair path (M8) as well as a trap for the oracle's scoping
-  (§3.3.4). Tracked as G16 and KBR-7.
+  (§3.3.4). Tracked as G16 and KBR-7. The declaration is now per-model —
+  `upstream_wire_is_messages_api_for_model(model)` mirrors `translate_to_upstream`'s own routing,
+  and the bare property reports the adapter's default (Chat Completions) route — and both bridge
+  repair sites read it from the `cc_request` the adapter routes on. The hook-level honesty guard
+  landed with the fix; §6.2.3 records what it does and does not prove. §3.3.4's decision stands
+  regardless: the oracle still selects on the observed shape, because an oracle must not ask the
+  code under test what shape it emitted.
 
 ### 4.5 Accepted residual risk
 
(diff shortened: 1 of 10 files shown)


shelpuk:
## Addressing the review's two suggestions

**`tests/test_wire_shape_honesty.py:336` — hand-written Converse fixture. Valid; fixed in 61c7500.**

The point is sharper than "low confidence" suggests. `WireShape.OTHER` is the classifier's *fallback*, so almost any dict satisfies it — meaning `test_classifier_reports_other_for_a_converse_body` and its tools-less sibling would both have stayed green through a typo that made `_CONVERSE_BODY` not resemble Converse at all. Of the classifier's known-positive controls, the one guarding the fallback was the one that could not fail.

The coverage itself was not missing: the registry sweep already runs the real `BedrockAdapter` body through the same classifier, so a genuinely mis-shaped classifier would go red there. What was missing was anything tying the fixture to that reality.

Added `test_the_converse_fixture_matches_what_bedrock_actually_emits`, which asserts the fixture and the body `BedrockAdapter.translate_to_upstream` actually emits agree on the two axes the classifier reads — presence of a top-level `tools`, presence of a top-level `system` — and that both classify alike. A typo in the fixture now fails loudly instead of silently.

**`src/kitty/providers/base.py:301` — `Args` type versus `.get("model", "")`. No change; I believe this is a non-finding.**

The parameter is annotated `model: str` and the `Args` entry documents it as the model name "exactly as `translate_to_upstream` will read it — unnormalized". Both call sites pass `cc_request.get("model", "")`, which returns `str` on every path including the missing-key one — the default is `""`, not `None`. So annotation, docstring and call sites agree, and there is no `str | None` to document.

The `.get` form is deliberate for a different reason, recorded in `_upstream_body_for`'s docstring: a subscript would raise `KeyError` inside the branch that exists to recover from an upstream rejection, converting a recoverable 400 into a 500. Matching the adapter's own read — same dict, same key, same default — is what makes the declaration and the emitted body agree by construction.

If the intent was that the docstring should state the empty-string default explicitly, say so and I will add it.

---

**Note on `ci-required`.** It is red here because `review-scripts` fails, and that failure is **pre-existing on `main`**, not introduced by this PR. `main` at `5687cf7` — the commit this branch was cut from — fails the identical assertion:

```
FAIL: GitignoredDocumentsAreNotPromisedTests.test_the_imperative_rule_recognises_what_it_forbids
.github/review/tests/test_review_scripts.py:13679
AssertionError: '.system_design' not found in {...}
```

The test asserts `.system_design` is git-ignored, but those design documents are tracked — `.gitignore` lists `/.requirements/` and not `.system_design`. Every other check on this PR passes, including the full matrix on Python 3.10–3.13.

I have deliberately not touched it. `.github/review/rules/python-tests.md` makes weakening an invariant guard to turn a diff green a **critical** finding, and this one needs its own ticket plus a decision about which side is actually wrong — the guard's expectation, or the `.gitignore`.

https://claude.ai/code/session_017EWo9i9PUfivaoe8sPL7K8

PR #52, opened 2026-09-07
Give every projection one vocabulary to be written in (KBR-25)

shelpuk:
Closes **[KBR-25](https://shelpuk.atlassian.net/browse/KBR-25)** — plan task **T-W2**, the input contract.

---

**What this is for.** Kitty Bridge's promise to the user is that it changes nothing in their request except what we deliberately decided to change — the model, mostly. Nothing proves that today. The test-suite design (KBR-2) specifies a "transparency oracle" that takes the request the coding agent sent and the bytes we actually put on the provider's wire, translates both into one neutral form, and fails on any difference nobody registered.

Those two things are in **different languages**: the agent might speak Anthropic Messages while the provider speaks Bedrock Converse. You cannot compare them as JSON. This PR defines the neutral language both get translated into.

**What it does not do.** It tests nothing about kitty yet. It is the dictionary; the eleven tasks behind it are the translators, the recorder, and the comparison itself. Nothing user-visible changes, and no product code is touched.

**Why it matters commercially.** This is the foundation for catching a whole class of bug that is currently invisible: a provider adapter that quietly drops a tool description, rewrites a token limit, or sends a request to the wrong Azure deployment. Those degrade output quality in ways that look like the model being bad rather than the bridge being wrong.

**Cost.** The plan sized this at half a day; it took about a day and a half. Two things moved into it during design review that the plan had not anticipated (below). Because this task sits at the head of every dependency chain, **every milestone moves one day** — the programme goes 17 → 18 days. The chain table in the plan is recomputed rather than left stale.

---

## What is in the change

| File | |
|---|---|
| `tests/harness/contract.py` | The contract — types, protocols, the totality rule, the path vocabulary |
| `tests/harness/test_contract.py` | 106 tests, including the falsification set |
| `.system_design/TEST_SUITE.md` | §3.3.1 corrected; §3.3.1a and §3.3.1b added |
| `.system_design/TEST_SUITE_IMPLEMENTATION_PLAN.md` | T-W2 re-scoped and re-sized; schedule recomputed |

No file under `src/kitty/` is touched.

### The load-bearing decision

The ticket names the falsification case: *"a stub reader that drops an unknown key fails it."*

The obvious implementation — "check nothing was left unclassified" — **cannot catch that reader.** A reader that drops a field leaves nothing behind to find, so the check passes over the exact defect it was written for. Totality is only decidable against the original body, so a projection now declares what it *accounted for*, and the check compares that against the source.

### Two additions from design review

Both are here for the same reason the ticket gives for putting `CapturedRequest` here rather than in T-W4 — one contract, one reading:

- **A path vocabulary and its matcher.** The next task (KBR-26) must record, for all 35 registered mutations, which field each one touches; the oracle must emit the same strings when reporting a difference. Neither can define that vocabulary without the other agreeing.
- **Cross-format normalisation rules.** Gemini calls the assistant `model`; OpenAI puts a tool result in its own message where Anthropic nests it inside a user turn. Six readers written by six authors would each choose differently, and the oracle would then report *our own inconsistency* as a broken promise — a false alarm on every Gemini request, which is the fastest way to get a safety check ignored.

### Verified, not asserted

Plan §1.4 exists because four earlier harness designs would have passed while proving nothing. So five deliberately broken versions of the check were built, and all five are killed by the test set:

| Sabotage | |
|---|---|
| Only checks for unclassified fields | killed |
| Rejects everything, always | killed |
| Checks requests, silently skips replies | killed |
| Credits a nested field to its parent | killed |
| Reports only the first missing field | killed |

The last three came from code review — the set that shipped first missed them.

### Facts checked against vendor documentation

- Bedrock Converse tool results carry `json` (the common case), `document`, `video`, `searchResult`; Anthropic carries `document` and `search_result`. The original shape held only text and images and would have dropped all of it **silently and identically on both sides** — invisible to the check.
- Gemini's `functionCall`/`functionResponse` carry **no id**, so tool-call ids are optional and pairing falls back to name and order. Ollama is recorded as unverified and handed to KBR-38.

---

## Review trail

Three `system-design-reviewer` rounds on the requirements, two `code-reviewer` rounds on the code, and two automated review rounds on this PR. Notable outcomes:

- The reviewer showed the original scoping decision on two recorder fields was wrong — four tasks consume them, and my split would have forced them all onto a different type. Reversed.
- I declined one proposed fix and said why: it would not have fixed the case it was raised for. The reviewer agreed on re-review.
- The final blocking finding was that my own matcher fix left the **design document** describing behaviour the code no longer had — the precise failure this task exists to prevent. Corrected, along with an anchoring warning that follows from it and is now carried by KBR-26 and KBR-53.

## Verification

- `ruff check .`, `lint-imports`, `mypy src/kitty` — pass
- `mypy tests/harness/contract.py` — clean (CI does not gate `tests/`)
- Full suite — green on **Python 3.10, 3.11, 3.12 and 3.13** in CI (~19 min each), plus `ci-required`. Locally the suite ran clean at the 93-test revision (3346 passed, 37 skipped, against a 3257-passed baseline on `main`); later local re-runs were killed by machine memory pressure, so CI across all four versions is the authoritative result.
- Bare-`pytest` collection verified in the CI form, since this repo has previously had a green local run and a red CI from that difference.

## Handed on, deliberately

- The 35-row assignment goes to **KBR-26**, reading the register's own data. A copy inside this task would be a second source of truth that stays green while the register moves.
- The tool-result merge rule goes to **KBR-34**: proving it needs two independently written readers, and a contract-level test would only show the author can copy.
- Route accounting splits between **KBR-36** (reader) and **KBR-52** (assertion).
- Body-borne credentials in captures, and whether Ollama's sampling options widen the canonical set, are recorded as open scope questions rather than decided here.

[KBR-25]: https://shelpuk.atlassian.net/browse/KBR-25?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.system_design/TEST_SUITE.md
+++ b/.system_design/TEST_SUITE.md
@@ -348,8 +348,9 @@ from it and carried separately, so a changed media type is its own delta. Gemini
 `fileData.fileUri` has no bytes: `digest` is then absent and `ref` holds the URI. Unpinned, the
 Messages reader and the Chat Completions reader would produce different digests for one image.
 
-The contract lives in `tests/harness/contract.py` (T-W2), which is also the home of T-W4's
-recorder, T-W5's proxy fixture, T-W6's corpus loader and T-W8's bridge fixture.
+The contract lives in `tests/harness/contract.py` (T-W2). The **package** `tests/harness/` is the
+home of T-W4's recorder, T-W5's proxy fixture, T-W6's corpus loader and T-W8's bridge fixture, each
+in its own module beside it — `contract.py` itself captures nothing and reads nothing.
 
 **Unknown fields fail closed.** Each reader must classify **every** key in the body into exactly
 one of: mapped to the envelope, mapped to the conversation, or residual. A non-empty `residual`
--- a/.system_design/TEST_SUITE.md
+++ b/.system_design/TEST_SUITE.md
@@ -478,6 +478,12 @@ agree on a canonical form. They are six separate tasks, so the agreement is part
   building an escape out of the residual defeats the escape. A value mapped to `other` has been
   seen and classified — it is accounted for. The residual means only *nobody has looked at this*.
 
+  **The pairing is enforced, both ways.** `other` without `stop_reason_raw` is rejected, because a
+  reader that maps both `SAFETY` and `RECITATION` to a bare `other` has discarded exactly what
+  T-D10 needs; and a `stop_reason_raw` beside a canonical reason is rejected as a stale leftover.
+  An invariant stated only in a docstring is a comment, not a rule — the same posture the closed
+  vocabularies take.
+
   **`usage` is carried but excluded from the diff**: it is provider-reported, never agent-supplied,
   so a difference carries no I1 information.
 
--- a/.system_design/TEST_SUITE.md
+++ b/.system_design/TEST_SUITE.md
@@ -319,6 +319,14 @@ and catches drops; the **path-keyed** residual covers nesting and fails closed, 
 because Gemini puts every sampling parameter under `generationConfig` and Converse nests
 `inferenceConfig` and `toolConfig`.
 
+> **The boundary, stated so nobody over-reads a green run.** `consumed` holds *top-level* keys, so
+> a reader that claims `generationConfig` and **silently drops** `topK` inside it **passes**
+> `verify_total`. Catching that would require the contract to walk the body itself — making it a
+> second reader, which the independent-oracle rule forbids. What closes it instead: each reader's
+> own L1 tests against its format's published examples (§7.4), and **T-D8**, which owns "residual
+> empty across the whole corpus" for all seven readers. T-W2 pins this boundary with its own test,
+> so it stays a decision rather than an assumption.
+
 **Optional ids, because two formats have none.** Gemini's `functionCall`/`functionResponse` carry
 no id; pairing there is by tool name and the k-th unanswered call of that name in the most recent
 assistant turn. A required id would force those readers to synthesise one and show a delta on every
@@ -462,11 +470,16 @@ agree on a canonical form. They are six separate tasks, so the agreement is part
   `tool:<name>`. This is the one deliberate exception to keying `extra` by the wire key, because
   four spellings name one concept.
 - **On the response direction**, `stop_reason` is `end_turn` · `max_tokens` · `stop_sequence` ·
-  `tool_use` · `error` · `other`, where `other` carries the original string in
-  `residual["stop_reason"]` — Gemini adds `SAFETY` and `RECITATION`, and a closed set with no escape
-  would fail the run on a legitimate safety-blocked reply. **`usage` is carried but excluded from
-  the diff**: it is provider-reported, never agent-supplied, so a difference carries no I1
-  information.
+  `tool_use` · `error` · `other`, where `other` keeps the wire's own string in
+  **`Reply.stop_reason_raw`** — Gemini adds `SAFETY` and `RECITATION`, and a closed set with no
+  escape would fail the run on a legitimate safety-blocked reply.
+
+  **Not in the residual**, and the reason generalises: a non-empty residual *fails the run*, so
+  building an escape out of the residual defeats the escape. A value mapped to `other` has been
+  seen and classified — it is accounted for. The residual means only *nobody has looked at this*.
+
+  **`usage` is carried but excluded from the diff**: it is provider-reported, never agent-supplied,
+  so a difference carries no I1 information.
 
 Then write one **hand-written reader per wire format** — Anthropic Messages, Chat Completions,
 OpenAI Responses, Gemini, Bedrock Converse, Ollama `/api/chat` — each written directly against
--- a/.system_design/TEST_SUITE.md
+++ b/.system_design/TEST_SUITE.md
@@ -287,24 +287,61 @@ the whole request and accounts for every field:
 Request                          -- the complete request; nothing is dropped silently
   envelope:     Envelope         -- routing and control
   conversation: Conversation     -- semantic content
-  residual:     {key: value}     -- anything the reader did not account for
+  residual:     {path: value}    -- anything the reader did not account for
+  consumed:     {key}            -- top-level body keys the reader DID account for
+  source:       {key: value}     -- the mapping the reader parsed
 
 Envelope
   model, stream, store, and every other control field the format defines
-  (Bedrock's modelId and Responses' store live here too)
+  in `extra`, keyed by the wire key
+  (Bedrock's modelId normalises onto `model`; Responses' store lives here too)
 
 Conversation
   system:   ordered text parts
-  turns:    ordered [ Turn(role, parts) ]
+  turns:    ordered [ Turn(role, parts) ]        -- role is `user` or `assistant`, only
   tools:    ordered [ ToolDecl(name, description, schema, strict) ]
-  sampling: { max_tokens, temperature, top_p, stop, seed, ... } -- declared, may be absent
+  sampling: a CLOSED set of fifteen canonical keys (§3.3.1b) -- declared, may be absent
 
 Part = Text(str)
-     | ToolUse(id, name, arguments)
-     | ToolResult(tool_use_id, content, is_error)
-     | Thinking(text)
-     | Image(digest)
+     | ToolUse(name, arguments, id?)
+     | ToolResult(content, tool_use_id?, is_error)   -- content: [ Text | Image | Json | Opaque ]
+     | Thinking(text, signature?)
+     | Image(digest?, media_type?, ref?)
+     | Json(value)
+     | Opaque(kind, digest?)
 ```
+
+**`consumed` is why a dropped key is detectable.** A reader that *drops* an unknown key produces
+an **empty** residual, so "the residual must be empty" would pass it — and T-W2's own falsification
+case is a stub reader that drops an unknown key. Totality is decidable only against the source
+body, so the projection records what the reader claimed to handle. `consumed` covers top-level keys
+and catches drops; the **path-keyed** residual covers nesting and fails closed, which matters
+because Gemini puts every sampling parameter under `generationConfig` and Converse nests
+`inferenceConfig` and `toolConfig`.
+
+**Optional ids, because two formats have none.** Gemini's `functionCall`/`functionResponse` carry
+no id; pairing there is by tool name and the k-th unanswered call of that name in the most recent
+assistant turn. A required id would force those readers to synthesise one and show a delta on every
+tool turn.
+
+**`ToolResult.content` is wider than text and images**, because Converse's `toolResult.content`
+carries `json` (the common case), `document`, `video` and `searchResult`, Anthropic's carries
+`document` and `search_result`, and Gemini's `functionResponse.response` is a bare struct. `Json`
+carries structured results; `Opaque(kind, ...)` keeps the rest **detectable** without modelling six
+vendors' block zoos, with `kind` a canonical snake_case name rather than the wire's spelling.
+
+**An empty block is a part with an empty string, never nothing.** P5e injects an empty `thinking`
+block and P8 an empty `reasoning_content`; P8's trigger is conditional and *inferred*, so §3.3.2
+assertion 2 needs its absence to be observable. `Thinking.signature` carries what M8's carrier
+repair manipulates.
+
+**`Image.digest` is the lowercase hex SHA-256 of the decoded bytes**, with `media_type` excluded
+from it and carried separately, so a changed media type is its own delta. Gemini's
+`fileData.fileUri` has no bytes: `digest` is then absent and `ref` holds the URI. Unpinned, the
+Messages reader and the Chat Completions reader would produce different digests for one image.
+
+The contract lives in `tests/harness/contract.py` (T-W2), which is also the home of T-W4's
+recorder, T-W5's proxy fixture, T-W6's corpus loader and T-W8's bridge fixture.
 
 **Unknown fields fail closed.** Each reader must classify **every** key in the body into exactly
 one of: mapped to the envelope, mapped to the conversation, or residual. A non-empty `residual`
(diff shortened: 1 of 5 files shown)


shelpuk:
Both findings were right and are fixed in d9bbc46. Thank you — the first in particular is a self-contradiction I introduced and did not spot across five review rounds.

## Warning — the `other` escape contradicted the totality rule

Correct, and the escape defeated itself exactly as described. I had specified that an unmapped wire value projects as `stop_reason: "other"` with the original string in `Reply.residual["stop_reason"]` — while `verify_total` fails the run on *any* non-empty residual. A reader following the contract literally would have failed **every** safety-blocked Gemini reply, which is the precise failure mode the escape was added to prevent.

The deeper error is that the residual was the wrong home on the contract's own terms. A value mapped to `other` **has been seen and classified** — it is accounted for. The residual means only *nobody has looked at this*. Putting a classified value there conflated the two.

Fixed by adding `Reply.stop_reason_raw`, which carries the wire's own string when `stop_reason` is `other`. A new test asserts the escape does not fail the run and that a Gemini `SAFETY` stays distinguishable from a `RECITATION`. `TEST_SUITE.md` §3.3.1b is corrected, and the generalisation is recorded there:

> Building an escape out of the residual defeats the escape.

## Suggestion — the nested-drop test was misnamed

Also correct, and I would rank it higher than a suggestion: **a falsification test that advertises cover it does not provide is worse than a missing one**, because it stops anyone looking again. The stub in that test residualises the nested key; a reader that genuinely *drops* it still passes.

You correctly identified the gap as design-intentional. Rather than only renaming, I have done both halves:

1. Renamed to `test_a_nested_key_the_reader_could_not_classify_fails_closed`, which is what it actually proves.
2. Added `test_a_nested_key_the_reader_silently_drops_is_a_known_limit`, which **asserts the nested drop passes** and documents why: catching it would mean the contract walking the body itself, making it a second reader — which the independent-oracle rule (§3.3.1) forbids. What closes it instead is each reader's own L1 tests against published examples (§7.4) and **T-D8**, which owns "residual empty across the whole corpus" for all seven readers.

That second test fails the day someone extends totality downward, which is the intent. It is a boundary marker, not an endorsement. The same boundary is now stated in `TEST_SUITE.md` §3.3.1 as a block quote, so a dependent task's author cannot over-read a green run.

## Verification

95 tests (was 93). `ruff check .` and `mypy` clean. All five falsification mutants re-verified as killed — residual-only, always-raises, reply-blind, nested-residual-credits-parent, reports-only-the-first-key.

shelpuk:
All four fixed in b25c7d6. None were blocking, but each was cheaper to fix than to carry, and one is a factual error I had propagated into three places.

## Warning — closed vocabularies declared but not enforced

Right, and the inconsistency was mine: `Turn` validated roles and `Conversation` validated sampling keys, while `STOP_REASONS` and `TOOL_CHOICE_VALUES` were declared closed and checked nowhere. **A vocabulary declared closed and enforced nowhere is a comment, not a rule** — and this contract's whole purpose is stopping six independently written readers from disagreeing. Passing a wire spelling straight through is exactly what they would have done.

- `Reply` now rejects a `stop_reason` outside the set, with the error naming the `other` + `stop_reason_raw` escape so the fix is obvious at the point of failure.
- `Envelope` now rejects an `extra["tool_choice"]` outside `auto` / `any` / `none` / `tool:<name>`.

`extra` stays open otherwise, deliberately — it holds whatever control fields a format defines, keyed by the wire key. `tool_choice` is the one entry with a canonical *value*, so it is the one entry worth checking. There is a test asserting other `extra` keys are **not** validated, so that boundary is deliberate rather than incidental.

## Warning — mask guard asserted 4 of the 5 headers its own docstring names

Correct. `proxy-authorization` was missing — and it is the one a containment recorder meets first, on the CONNECT legs (§5.2.1). Added to the assertion, plus a separate test that the mask **actually reaches it**, since membership in the set is not proof the redaction fires.

## Suggestion — "Ten tasks depend on it" lists eleven

Not a miscount in the list; the **count was wrong**, and I had propagated it into the plan, the PR description and the Jira comment. Counted from the issue links: KBR-26, -27, -32, -33, -34, -35, -36, -37, -38, -39, -51 — **eleven**. Corrected in the module docstring, `TEST_SUITE_IMPLEMENTATION_PLAN.md` §14.4, the PR description and the Jira ticket.

## Suggestion (low confidence) — import guard missed dynamic `src.kitty`

Confirmed and worth taking despite the confidence flag: `import_module` and `__import__` matched only the bare package, so `import_module("src.kitty.bridge.server")` would have slipped through. Widened, with both forms added to the positive control — which is what makes the gap visible next time.

## Not changed

The PR description's stale test count is fixed (93 → 101), along with the verification section: later local full-suite runs were killed by machine memory pressure, so **CI across Python 3.10–3.13 is the authoritative result** rather than my local run. Said plainly rather than quoting a local number I can no longer reproduce on this machine.

## Verification

101 tests (was 95). `ruff check .`, `mypy src/kitty` and `mypy tests/harness/contract.py` clean. All five falsification mutants re-verified as killed.

shelpuk:
Fixed in 0c74350.

Correct: `REDACTED_HEADERS` held seven entries while its docstring explained five. The unexplained pair was `cookie` and `set-cookie`.

They are **precautionary, not observed** — no adapter authenticates by cookie today, so nothing in the register or §7.2 exercises them. I kept them rather than trimming to match the prose, and said why: a session cookie is a credential, listing one nobody sends costs nothing, and omitting one somebody later starts sending leaks it into every CI log. `CapturedReply` is the likelier carrier of the two. The docstring now separates the five carriers from the two precautionary entries.

The test also now asserts the set **exactly** rather than as a subset. The subset form is what allowed the drift — it could not notice an entry the docstring did not explain. Exact assertion closes it in both directions.

101 tests, ruff and mypy clean.

shelpuk:
Fixed in 1b81f03.

Correct, and the interesting part is that this is round 3's finding one site over. Round 3 was "the constant's prose disagrees with the constant"; I fixed the prose and the assertion and left a **third** copy of the same claim in the test's name and docstring. Two of three updated reads like a complete fix until somebody greps.

The test is renamed to `test_the_mask_set_is_exactly_five_carriers_plus_two_precautionary` and its docstring now states both groups — the five with a carrier the recorders will meet, and `cookie`/`set-cookie` as precautionary with the reason they are kept. It also records why the assertion is exact rather than a subset: the subset form could not notice an entry nobody had explained, which is how the drift opened.

101 tests, ruff clean. No behaviour change.

shelpuk:
Both round-5 findings fixed in f0491f6. You were right to re-raise them — the merge commit did not touch the harness files, so they were genuinely still standing rather than superseded.

Both are the same defect, and it is the third time this PR has been caught by it: **a rule written in prose and wired to nothing.** Round 2 was two closed vocabularies declared and unchecked; round 3 and 4 were a constant whose docstring disagreed with it; these two are an invariant living only in a docstring and a mask asserted for ten entries but exercised for four. Worth naming, because it is clearly my systematic blind spot on this change rather than four unrelated slips.

## Warning 1 — `stop_reason_raw`'s pairing was a docstring, not a rule

Correct. The field is documented as carrying the wire's value when `stop_reason` is `other` and `None` otherwise, and neither direction was checked. `Reply.__post_init__` now rejects both:

- **`other` with no raw value** — a reader that maps both Gemini's `SAFETY` and `RECITATION` onto a bare `other` has discarded precisely what T-D10 needs to tell them apart. The escape's whole value is the pairing.
- **A raw value beside a canonical reason** — a stale leftover from a mapping that succeeded.

The new rule immediately failed one of my own tests, which looped over every canonical reason constructing each one bare — including `other`. That is the guard working on its first run, and I have left the corrected control plus a test for each direction.

## Warning 2 — the mask asserted ten entries and exercised four

Correct, and the shape of the round-2 fix is why it recurred: I added `proxy-authorization` **one entry at a time**, which fixes the instance and leaves the class. Membership and lower-casing were pinned for all ten; only four were ever driven through `_redact_headers` / `_redact_query`, so six could have sat in the set and been silently unmasked.

Now swept **from the constants themselves**, both sets, so an eighth entry cannot ship asserted-but-unexercised. Each value is sent in **upper case**, which proves the match is case-insensitive per entry rather than inferring it from the constant happening to be lowercase.

Verified by mutation rather than asserted: making `x-goog-api-key` silently unmasked is now **caught**. The hand-picked tests missed exactly that.

## Not raised, and I agree

The two items the merge brought onto this branch's HEAD — KBR-132's recorded `pytest.skip` and `TEST_SUITE.md` §8.1/§8.2 — are KBR-24's, not this PR's. No harness or product code here touches either.

## Verification

104 tests (was 101). `ruff check .`, `mypy src/kitty` and `mypy tests/harness/contract.py` clean. The five falsification mutants re-verified as killed, plus the new mask mutant. The pairing rule is recorded in `TEST_SUITE.md` §3.3.1b, since that is the file the dependent tasks read.

shelpuk:
Fixed in a558d58. Correct, and it is a regression I introduced *one commit ago* while fixing your round-5/6 finding — worth saying plainly rather than filing as a fresh defect.

## The finding

The header sweep asserts the mask token is present. The query sweep asserted only that the secret was gone and that an innocent parameter survived. So a `repr` that dropped the credential pair outright, or emptied it to `key=`, would have passed **while looking like proof of masking**.

That is precisely the hazard round 1's `test_the_masked_repr_still_shows_that_a_credential_was_present` exists to close on the header side: masking to nothing hides a *missing-credential* bug exactly as effectively as no masking hides a present one. The query side had that assertion when it was a single hand-written test; rewriting it as a derived sweep fixed the coverage gap and dropped the stronger assertion on the way through.

**Fixing coverage by rewriting a test can lose an assertion the old test had.** I gained six entries and lost one property, and the net was a weaker check on a credential path.

## The fix

The query sweep now asserts the pair survives **with the mask as its value**, matching the header sweep. Verified by mutation rather than by inspection, in both shapes the finding names:

```
KILLED - pair dropped entirely
KILLED - pair emptied to `key=`
```

## On the withheld suggestions

Noted that two suggestion-level findings were withheld this round as out of scope from round 6 onward. I cannot act on findings I cannot see. If they are surfaced in a later round, or if a human reviewer wants them applied, I will take them — I would rather have them than not, given six of the eight findings so far have been worth fixing.

## Verification

104 tests. `ruff check .` and `mypy tests/harness/contract.py` clean. The five falsification mutants re-verified as killed, plus the header-mask mutant and both query-mask mutants. All 12 checks passed on the previous commit `f0491f6`, including Python 3.10–3.13.

shelpuk:
Both fixed in 5b2c1c3. Thank you for surfacing the two that round 7 withheld — both were worth having, which is why I asked.

## Suggestion 1 — a two-character string walked through the header-pair guard

This is the better of the two, and it is embarrassing in an instructive way. `tuple("ab")` is `("a", "b")`: **length 2**, so my length check passes it, and it is stored as a header named `a` with the value `b`. Silent shredding — through the guard I added in round 2 *specifically to stop silent shredding*, whose docstring says "Failing loudly is the point."

A string is a sequence, so no length check can catch this; only rejecting the type does. Now rejected, with two tests: the 2-character case a length check misses, and a longer string it would have caught anyway — the second being the control that shows the first is not redundant.

The lesson generalises past this line: **a guard against "wrong shape" that validates only *size* is blind to any type that happens to have the right size.** Same family as the mapping case that guard was originally written for.

## Suggestion 2 — the design docs homed four other tasks' deliverables inside `contract.py`

Correct, and the PR contradicted itself three ways. Both design documents said `tests/harness/contract.py` "is also the home for T-W4's recorder, T-W5's proxy fixture, T-W6's corpus loader and T-W8's bridge fixture", while the module's own docstring says "It captures nothing — the recorders do that" and `__init__.py` says those tasks "land here alongside it".

The **package** is the home; `contract.py` defines shapes and rules and reads nothing. Corrected in both documents — which matter more than the docstrings here, since the dependent tasks read `TEST_SUITE.md` and not this module.

## Also corrected

The PR description's test count was stale at 101 against 106 at head. You recorded it in prose rather than raising it because no file line anchors it; that was the right call, and it is now fixed.

## Verification

106 tests. `ruff check .` and `mypy tests/harness/contract.py` clean. The five falsification mutants re-verified as killed, plus the header-mask and both query-mask mutants. All 12 checks passed on `a558d58` including Python 3.10–3.13; the new `isinstance(entry, str | bytes)` is PEP 604 syntax valid from 3.10, and the 3.10 job will confirm it definitively.

shelpuk:
Both fixed in 2e7eabd. The warning is a regression I introduced in the round-8 fix, and you are right that it made the input *worse* than before that commit.

## Warning — the new check read `headers` twice

`any(isinstance(entry, ...) for entry in self.headers)` consumed the input, and `tuple(tuple(h) for h in self.headers)` then saw nothing — so a generator was stored as **no headers at all**, silently, from the guard whose docstring says failing loudly is the point. A recorder streaming its headers is the realistic source, and T-W4 is the task most likely to hand one over.

Fixed by materialising once, up front.

Worth naming the pattern, because this is its **third** appearance on this branch: a guard against silent loss that itself loses silently. Round 2's mapping shredding, round 8's two-character string, and now this. The shape is constant — **the check and the store disagreeing about what they are looking at.** Each fix closed one instance and left the shape intact. Materialising once closes the shape, because there is now a single value that both the checks and the store read.

## Suggestion — `CapturedReply`'s copy was untested

Correct, and the two findings are the same finding. The guard was **hand-copied** into both capture types, so the bug existed twice, and `CapturedReply`'s copy had no test on any of its three rejection paths — a defect found in one was silently a defect in the other.

They now share `_normalised_headers`. That is the actual fix: it makes a correction apply to both *by construction*, rather than by my remembering to paste it twice. The reply side is parametrised over all three rejection paths plus the generator case, so the shared behaviour is pinned from both ends rather than assumed from one.

## Verified by mutation, not inspection

Reintroducing the exact double-read shipped in 5b2c1c3:

```
KILLED - headers read twice (the round-8 regression)
2 failed, 109 passed
```

Two failures, one per capture type — where before this commit it failed none.

## Verification

111 tests (was 106). `ruff check .`, `mypy src/kitty` and `mypy tests/harness/contract.py` clean. The five falsification mutants re-verified as killed, plus the header-mask, both query-mask, and this new double-read mutant.

shelpuk:
Both fixed in 4d76e51. The first is a **real credential leak**, not a tidiness point, and I reproduced it before touching anything:

```
bytes header name accepted: ((b'Authorization', 'Bearer sk-LEAKED'),)
SECRET IN REPR: True
```

## Warning 1 — a bytes header name walked past the mask

Exactly as described. The pair guard checked a pair's **length** and not its element **types**, so `(b"Authorization", "Bearer sk-...")` was accepted — and `b"authorization"` is not in a set of `str`, so redaction missed it and the token was rendered in full into every log and assertion diff.

Fixed by requiring both elements to be `str`, covered on the name *and* the value so the check cannot be half-applied.

This is the third instance of one shape, and I think it is now worth stating as the through-line of this whole review: **rounds 8, 9 and 10 are all a guard that validates the size of a thing while ignoring what the thing is made of.** A 2-character string has length 2. A generator has entries until you read it twice. A bytes name has length 2 and lowercases fine. Each fix closed an instance; none closed the shape.

## Warning 2 — the `Part` union was declared closed and enforced nowhere

`Turn(role="user", parts=("not a part", 42))` constructed happily. Every other closed set here is checked at construction — `ROLES`, `SAMPLING_KEYS`, `STOP_REASONS`, `TOOL_CHOICE_VALUES` — so this was the odd one out, and "closed" was doing no work.

**I fixed every field that declares a closed set, not only the one you reported**: `Turn.parts`, `Reply.parts`, `ToolResult.content`, `Conversation.system`, `.turns`, `.tools`. Rounds 3, 4, 8 and 9 were each a version of "you fixed the instance, here is the same defect one site over" — six sites fixed individually would have been four more rounds.

`RESULT_PART_TYPES` is a side benefit: the contract's claim that a tool call cannot nest inside a tool result is now *checked* rather than only written in a docstring.

Both go through one helper that materialises its input once, so round 9's one-shot-iterable defect cannot reappear across six new call sites.

## Verification

119 tests (was 111). `ruff check .` and `mypy tests/harness/contract.py` clean. Both holes re-checked as closed after the fix; the five falsification mutants, the header-mask, both query-mask and the double-read mutants all still killed. All 12 checks passed on `2e7eabd`, including Python 3.10 — which also settles the `isinstance(x, str | bytes)` PEP 604 question I flagged in round 8.

## On the withheld suggestions

Round 10's footer again records 2 withheld. I have asked twice; I will not ask a third time, since the convention is clearly deliberate. Noting only that round 8's two withheld items both turned out to be worth fixing, one of them the `tuple("ab")` corner that led directly to this round's leak.

PR #54, opened 2026-09-11
Split the leak guard by evidence, and stop blinding the automated reviewer

shelpuk:
Closes KBR-128.

This PR fixes two silent failures in the machinery that reviews our pull requests, and one live defect on the public README.

**The automated reviewer has been working blind since PR #43.** When we started committing our design documents, five configuration documents kept telling the AI reviewer that those documents are not in the checkout — in so many words: *"do not look for a design document, do not report a missing one as a finding, and do not claim to have read one."* Every review since has judged changes against the README alone, ignoring the internal specification it was standing on. Nothing went red, because the failure mode of a blinded reviewer is silence — it looks exactly like a clean diff.

The reviewer is now told the design documents are present, that they are the internal specification alongside the README as the user-facing one, and to read the section covering the area a change touches. It is also told that a **stale** design document is itself a finding — a design left behind by the change that invalidated it is a wrong docstring one level up, and worse, because the next author reads it as the spec.

**Our leak guard was checking one directory out of the whole repository.** This is a public repository; a guard forbids tracked files from naming a private project, and its failure message states that policy as absolute. It was only ever applied to `.github/`. Its own comment predicted this and asked to be extended; the day came and nothing failed, because the omission sat in a hand-written list rather than a walk.

**The README's CI badge and logo pointed at this repository's former name.** They work today only because GitHub redirects renamed repositories. Fixed.

## The decision worth your attention

The ticket recommended simply extending the guard to more directories. That was tried first and **measured**, and the measurement changed the plan.

The guard has two independent rules, and they do not generalise alike:

| Rule | Hits outside `.github/` | Real leaks |
|---|---:|---:|
| Repository-name rule (`Owner/name`) | 2 | 0 |
| Issue-key rule (`ABC-123`) | 341 on 309 lines, in 38 forms | **0** |

<sub>Measured at `acd9e9b`, 2026-09-11. Dated because it moves: merging `main` took it to 344/312/39 within hours without touching anything the rule cares about. The distribution is the argument and does not change; `SWEPT_DIRS` carries the command to re-derive the number.</sub>

The issue-key rule matches a *shape*, never a name — that is its whole design, and it is why it works inside `.github/`. Everywhere else this codebase writes hundreds of tokens of the same shape that are not issue keys at all: acceptance-criterion IDs, test scenario IDs, model names like `GPT-5`, protocol tokens like `HTTP-413`. No pattern separates those from a real tracker key. The documented consequence of a guard that cries wolf is that people delete the *prose* rather than fix the pattern.

So the two rules are split by evidence. The repository-name rule — the one that catches what actually matters, a private repository's real name — now sweeps **every tracked file**. The issue-key rule keeps the scope where its premise holds and **says so in its failure message**, so a green run is not read as a statement about the rest of the tree. What that leaves uncovered is stated plainly rather than left to be discovered.

One narrowing is added deliberately and recorded as a narrowing: our own tracker prefix is allowed. The alternative was rewriting every one of those citations — they run to three figures and grow with each ticket — when they carry the traceability our test-suite design is built on, for a key already published alongside its tracker URL. **Not a silent trade — the constant explains it, and the controls exercise it in both directions.**

## Still open, and deliberately not decided here

Our design documents cover the bridge's test suite, not the review system. The review workflow already records that *"nobody has yet decided where review-workflow decisions belong."* Rather than invent a home for that as a side effect, the reasoning for these decisions goes where this repository already keeps review-system design — the class docstrings and `.github/review/README.md`. Worth a separate decision.

## What changed

- **Five documents plus the rule selector's comments** corrected. Prose lifted from `16782c7`, written and reviewed on another branch and reverted there to respect a scope decision.
- **Two guard rules imported** so those sentences cannot return on the next upstream sync, anchored so the sweep's line-joining cannot manufacture a false match.
- **A new positive guard** asserting the corrected instructions are *present*. A sweep proves a sentence is gone and can never prove one is there — and the defect here was an absence.
- **The leak guard split**, and a latent defect fixed along the way: its two entry points each restated both rules under a docstring claiming they could not drift, so every control was exercising a second copy of the rule the sweep ran.
- **The gitignored-path guard** swept three documents and not the two rule files, both of which instruct the reviewer. It now walks the rules directory.
- **One stale sentence stood in two docstrings** in the same file, months apart, the second a copy of the first. That is the argument for a guard over a one-time edit, and both are now inside the sweep.

## Verification

- Review suite: **620 tests, green**, up from 606. None removed.
- Full repository gate after merging `main`: **3497 passed, 2 skipped**. `ruff`, `import-linter`, `mypy` clean.

<sub>Both figures are as of `c2c0555`. They rot on the next commit, which is the point the automated review made below — and the reason the same figures were taken out of the code's prose in that commit.</sub>
- Every new control was verified **by mutation**, not by inspection. Eleven mutations each turn the suite red and name the control that catches them: empty `SPECIMENS`; pad `ALLOWED` with duplicates; delete the packaging near-miss; swap it for a different real paragraph keeping the count; delete the prompt's "is tracked" row; duplicate a `why` to pad the row floor; unmark the marked code line; narrow the rule's verb set; empty the own-key allowance; drop the scope sentence from the failure message; narrow the document sweep back to three files.

The change went through two rounds of design review and two of code review before landing. Three defects it introduced were caught and fixed in the second commit — each of them the same shape as the defect this PR exists to repair: a claim that stopped being true with nothing to say so. They are written up in the commit message rather than quietly corrected.

https://claude.ai/code/session_01JkxLUaQYgWJq9qVeGZSz7x


shelpuk:
## Status: `ci-required` green, one check failing for a reason outside this change

`ci-required` passes — all four `test` matrix jobs, `review-scripts`, `review_replies`, CodeQL and both `Analyze` jobs. `main` is merged in (12 commits); no file overlap, but the tree-wide sibling sweep this PR adds means files landing on `main` are now inside a guard they were not inside before, so it was merged and re-run rather than assumed.

### The `review` check: not this change, and the notice is misleading

The provider returned `HTTP 402 — Insufficient Balance`. **The account is out of credit**, which needs an operator top-up.

The notice this workflow posted says the opposite — *"a problem with the workflow configuration"*, *"needs a fix in the workflow"*. It is not. Reproduced against the classifier:

| Payload | Classified |
|---|---|
| The repo's own `TestClassify` fixture | `exhausted` ✅ |
| The **real** payload — identical plus `"code":"invalid_request_error"` | `fatal` ❌ |

`FATAL_PATTERNS` is matched before `QUOTA_PATTERNS` and contains `invalid_request`, which the 402 body happens to carry, so `insufficient balance` is never reached. The test fixture is an *abridged* version of the real error and the abridged part is the part that changes the answer. Second consequence: `fatal` stops the provider fallback chain, so a spent balance on one provider blocks review entirely instead of failing over.

Filed as **KBR-145** (High) with the reproduction and acceptance criteria. **Deliberately not fixed here** — it is an independent defect in a different module, and fixing it would not have made this review run, since the balance really is spent. Folding it in would widen a PR that is already large and put an unrelated classifier change under the same review.

### Deferred, with reasons

1. **No `.system_design/` document for the review system.** `TEST_SUITE.md` covers the bridge's suite and never mentions the review system, and `claude-code-review.yml` already records that *"nobody has yet decided where review-workflow decisions belong."* Settling that as a side effect of this ticket would be the wrong way to settle it. The rationale for these decisions therefore went where this repository already keeps review-system design — the class docstrings and `.github/review/README.md`. Worth its own ticket.

2. **The issue-key rule still does not sweep `src/`, `tests/` or `.system_design/`.** That is the decision, not an omission — measured, argued in `SWEPT_DIRS`, and stated in the sweep's own failure message so a green run is not over-read. The uncovered case is named there rather than left to be discovered.

3. **`SIBLING` cannot see a slug split across a line.** Two live instances (`model_context_sync.py` and its test both break a raw-content URL at the owner/name boundary). Neither is a leak, both name this repository. Recorded in the class docstring rather than closed: the fix is a line-join over the widest sweep in the module, for a shape that has produced no leak.

### Self-review found one defect worth naming

The `SWEPT_DIRS` docstring recorded the measurement that justifies the narrowing as an exact figure — "341 tokens on 309 lines in 38 forms". Merging `main` moved it to 344/312/39 within hours, touching nothing the rule cares about. **A hand-typed figure nothing checks, in the module whose `RecordedTestCountTests` exists because quoted figures rot** — and it appeared twice, which is the duplication that has already bitten this file twice. It now appears once, as the shape of the result (which does not move) plus a dated observation and a command to re-derive it. The command was run, and the docstring says it prints the *raw* count before the constants filter it rather than implying the numbers match.

### Verification

620 tests in the review suite (from 606, none removed). `ruff`, `import-linter`, `mypy` clean. **Eleven mutations each verified red**, naming the control that catches them: empty `SPECIMENS`; pad `ALLOWED` with duplicates; delete the packaging near-miss; swap it for a different real paragraph keeping the count; delete the prompt's "is tracked" row; duplicate a `why` to pad the row floor; unmark the marked code line; narrow the rule's verb set; empty the own-key allowance; drop the scope sentence from the failure message; narrow the document sweep back to three files.

Four rounds of code review, all findings applied. Six of the defects caught across those rounds were introduced by this change's own fixes — each the same shape as the bug the PR exists to kill: a claim that stopped being true, with nothing to say so. That is the argument for the guards this PR adds, made on this PR.

shelpuk:
## Re: the automated review — finding accepted, and it was under-counted

> PR description claims 619 tests; actual count is 620. This is the precise rot the very `RecordedTestCountTests` class exists to prevent, and it occurs in the one place the guard does not run — the human-written PR description.

**Correct, and the framing is sharper than the one I wrote.** Fixed — and checking properly, there were **four** rotting figures in the description, not one:

| Claim | Was | Actually |
|---|---|---|
| Review-suite size | 619 | **620** |
| Full repository gate | 3378 passed | **3497 passed** (stale since `main` was merged) |
| Issue-key measurement | 341 / 309 lines / 38 forms | **344 / 312 / 39** |
| Citations that would need rewriting | 127 | **130** |

The reviewer saw the one figure it could check against the tree. The other three drifted for the same reason and nothing could have caught them.

### What I changed, beyond the number

Bumping 619 to 620 would have left the description rotting again on the next push, so the description now gets the same treatment the code got in `c2c0555`:

- The measurement table is **dated** (`acc: acd9e9b, 2026-09-11`) with a note that merging `main` moved it within hours, and points at `SWEPT_DIRS` for the command to re-derive it. A dated observation is a historical fact and cannot rot; a live claim can.
- The prose figures are gone in favour of the shape — "hundreds of tokens", "three figures and growing" — which is what actually carries the argument and does not move.
- The two verification counts are stamped as of `c2c0555` and say outright that they rot on the next commit.

### The honest reading of this

This is the **fourth** time this defect class has surfaced on this PR, and three of those were mine. The first was the bug the ticket describes. Then the `SWEPT_DIRS` docstring recorded its own justifying measurement as an exact figure and was falsified by a merge within hours. Then the count guard's own docstring said "two documents" while `QUOTING` held three. Now the PR description.

That is not four unlucky typos — it is one lesson arriving four times: **a precise figure written into prose is a claim with no guard behind it, and the only durable forms are "checked by a test" or "dated and explicitly historical".** The code now holds both forms deliberately. The PR description holds the second.

It also marks a real limit of what this PR builds, worth stating rather than leaving implied: `RecordedTestCountTests` covers the three tracked documents that quote the suite size, and nothing more. Pull request descriptions, commit messages and Jira comments are outside every guard in this repository and always will be. The automated reviewer reading the description against the tree is the only thing that closes that gap — which is exactly what happened here, on the PR whose purpose is to stop that reviewer working blind.

No code change was needed for this finding; the tree was already correct at 620 in all three guarded documents (`ci.yml`, `pyproject.toml`, `.github/review/README.md`), which is why the guard was green and only the unguarded prose was wrong.

PR #74, opened 2026-09-11
T-A3 — OpenAI Responses reader (KBR-35)

shelpuk:

Kitty Bridge sits on the wire between a coding agent and an LLM provider. Its commercial promise is that it does **not** secretly change what the agent asked for — it may swap the model and the credentials, but the conversation must arrive intact. Today nothing checks that claim.

Checking it needs a comparison the bridge cannot fake: read the agent's inbound request **and** the bytes that actually went upstream into one neutral form, then diff them. Those two bodies are often in *different* wire formats, so they cannot be compared as JSON — only as projections. Six such readers are planned, one per format.

**This PR delivers the reader for OpenAI Responses — the format Codex speaks.** It ships no user-facing behaviour and changes no production code. What it buys is the ability, once the oracle lands, to answer "did the bridge alter this request?" with evidence instead of assurance. It unblocks T-D5 (KBR-55) and T-D9 (KBR-58).

Two things worth the Product Owner's attention:

- **It found two register gaps that would have produced false alarms.** The mutation register lists every change the bridge is *allowed* to make; anything unregistered reads as a breach. Sixteen fields the Codex backend legitimately drops are claimed by no register row — so the first fidelity run would have accused the bridge of tampering it never did. Filed as **KBR-171**, linked as blocking KBR-55, not fixed here because the register is a different task's to edit. **KBR-149** already covered a sibling gap; I checked before filing rather than adding a duplicate.
- **It is deliberately written against OpenAI's published schema and forbidden from importing any bridge code.** A reader that asked kitty how to read a body would inherit kitty's bugs, and the whole check would prove only that kitty agrees with itself.

---

## What this is

Plan task **T-A3** ([KBR-35](https://shelpuk.atlassian.net/browse/KBR-35)), design `.system_design/TEST_SUITE.md` §3.3.1/§3.3.1a/§3.3.1b. Depends on T-W2 (KBR-25, **Done**).

`tests/harness/reader_responses.py` exposes `ResponsesProjection`, satisfying `contract.Projection` for `WireFormat.OPENAI_RESPONSES`: a `CapturedRequest` in, a `contract.Request` out — envelope, conversation, residual, consumed.

Written against `openai/openai-openapi` `openapi.yaml` **v2.3.0**, pinned in the module docstring. Fixtures are the eight `x-oaiMeta` request examples the spec itself carries, adapted only to substitute `example.test` for third-party hosts.

**Three files. No production code is touched.**

| File | |
|---|---|
| `tests/harness/reader_responses.py` | the reader |
| `tests/harness/test_reader_responses.py` | 176 tests |
| `.system_design/TEST_SUITE.md` | **+28 lines** — see below |

### The design-doc change, called out explicitly

This PR edits a **shared** design document, which is easy to miss inside a code commit. Two insertions in §3.3.1b, nothing rewritten:

1. **The merge rule's four clauses are an ordered pipeline** — so `ToolResult → user text → ToolResult` keeps wire order rather than hoisting results over history the bridge did not move.
2. **Tool-call arguments decode through one shared rule**, with the test for *why* an absent `arguments` is accepted while an absent tool `name` is not: can the projection represent the absence losslessly?

Both were ambiguities in rules **five other readers** will implement, and a turn-ordering disagreement between two readers reports a false delta on every subsequent turn. They lived in this task's requirements until the design reviewer confirmed they belong where the other authors will actually read them. The code pin for the decode is filed as **KBR-174**, blocking T-A2 (KBR-34).

## What it classifies

All **31** published top-level keys. All **31** item types — four modelled semantically (`message`, `function_call`, `function_call_output`, `reasoning`), 27 kept as detectable placeholders. Nine `tool_choice` forms. Both `input` shapes. Anything outside those closed sets residualises, which fails the run: adding a field to a wire format must force a decision, not widen a blind spot.

Three choices worth their reasons:

- `max_output_tokens` normalises onto `max_tokens` and `strict` stays a tri-state, because register rows **P14** and **P15** are written against exactly those. `input_text` and `output_text` project identically, which is what makes **P16** `NOT_PROJECTABLE` — `contract.py` names this module's L1 test as one of that row's two guards, and it is here.
- **A field a modelled type declares either projects or residualises.** Never coerced, never skipped. `str(7)` invents a value the agent did not send; skipping an unreadable reasoning entry performs the very loss the per-entry fan-out exists to expose when a *bridge* does it.
- **A top-level key stays in `consumed` when a value beneath it residualises**, or `verify_total` reports a drop — the wrong diagnosis for a reader that read the body fine.

## Verification

- **176 tests**; 564 across `tests/harness/` plus the layer modules.
- `ruff check`, `ruff format --check`, `mypy src`, `lint-imports` 4/4 — all clean.
- **14 deliberate defects injected into the reader; 14 caught.** Tests that have never failed prove nothing.
- Merged `origin/main` (14 commits, incl. KBR-144/146/147) and re-verified; no conflicts.

Two design-review rounds and three code-review rounds found **five real defects**, every one of which passed a green test run:

| Defect | Why it mattered |
|---|---|
| `TypeError` escaping five unguarded membership tests | a malformed *client request* would have been diagnosed as a defect in our own harness |
| A lifted-message residual overwrote one of two entries and mislabelled the other | silent evidence loss, in a tool whose job is preserving evidence |
| Tool arguments in the wrong wire form compared equal | hid a real class of wire-format breach |
| Reasoning entries vanished silently | **introduced by me** while fixing something smaller |
| An unnamed tool sat at `conversation.tools[]`, which `path_matches` treats as a wildcard | register rows would have matched it by accident |

Each is fixed, and each has a test that fails if it returns.

Reviews: `code-reviewer` **APPROVE**; `system-design-reviewer` — no blocking findings, all wording feedback applied.

Refs KBR-35 · files KBR-171, KBR-174

https://claude.ai/code/session_01J2sAEXym49SsSKWadFLDqf

[KBR-35]: https://shelpuk.atlassian.net/browse/KBR-35?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.system_design/TEST_SUITE.md
+++ b/.system_design/TEST_SUITE.md
@@ -596,6 +596,16 @@ agree on a canonical form. They are six separate tasks, so the agreement is part
   still projects, in the turn where it occurred* — M7 exists to drop orphans, so a reader that
   raised on one would fail instead of producing the delta that names it.
 
+  **The four clauses are an ordered pipeline.** `ToolResult` parts come first *within the turn the
+  **run** and **following-message** clauses build*, never as a re-sort after the same-role merge —
+  so `ToolResult → user text → ToolResult` projects as `[ToolResult, Text, ToolResult]`: two runs,
+  the first absorbing the text that follows it, then concatenated. Re-sorting afterwards would
+  hoist a result ahead of text the agent sent **before** it — moving history the bridge did not
+  move — and because paths are index-based the invented delta would land on every part of that
+  turn and every turn after it. The rule gives a run of results one home; it does not reorder
+  history. Clause 3 is not thereby idle: it governs the formats that carry text and results inside
+  **one message**, where the run has no natural boundary.
+
   A lift rule ("into the user turn that follows the assistant turn") does **not** work: the standard
   Chat Completions exchange ends `assistant(tool_calls) → tool → tool`, with no following user
   message at all. And because paths are index-based, any disagreement about turn boundaries reports
@@ -624,6 +634,24 @@ agree on a canonical form. They are six separate tasks, so the agreement is part
   compares its value whole. No reader emits `envelope.extra[<key>.<subkey>]`; nesting belongs to
   the residual (§3.3.1a). Six readers cannot quietly disagree about whether `thinking.budget_tokens`
   has an address of its own.
+- **Tool-call arguments decode through one shared rule.** Chat Completions and Responses both
+  carry them as a JSON *string*. Absent, `null`, empty or whitespace decodes to `{}`; a value that
+  is valid JSON but not an object, or not valid JSON at all, decodes to `{}` **and residualises at
+  that argument's own path**. Never raises.
+
+  **Why an absent `arguments` does not residualise though an absent `name` does**, the schema
+  requiring both: the test is whether the projection can represent the absence *losslessly*.
+  `ToolUse.arguments` is a mapping defaulting to empty, and `{}` is a true statement about the
+  call — seen and classified, therefore accounted for, the same ground on which `STOP_REASONS`
+  gives `other` its escape instead of the residual. `ToolUse.name` is a `str` with no such value:
+  `""` claims a tool *named* empty-string, and a call nobody can name cannot be paired or
+  addressed. Apply that test to every required field, not only these two.
+
+  Six readers cannot quietly disagree about what `arguments: ""` means, and it is not
+  hypothetical — `openai_subscription.py:OpenAISubscriptionAdapter._cc_to_responses` writes
+  `func.get("arguments", "")`. The residual *path* is format-specific and stays each reader's own;
+  only the decode and the fail-closed policy are shared. Tracked for pinning as code beside
+  `image_digest` — KBR-174.
 - **`tool_choice`** unifies four wire keys — CC/Messages `tool_choice`, Converse's
   `toolConfig.toolChoice`, Gemini's `functionCallingConfig.mode` — onto
   `envelope.extra["tool_choice"]`, with the **value** normalised to `auto` · `any` · `none` ·
--- a/tests/harness/reader_responses.py
+++ b/tests/harness/reader_responses.py
@@ -1,0 +1,1185 @@
+"""The OpenAI Responses wire projection — one of the six independent readers.
+
+`.system_design/TEST_SUITE.md` §3.3.1, §3.3.1a, §3.3.1b · plan task **T-A3** (KBR-35).
+
+Projects an OpenAI Responses request body into :class:`harness.contract.Request`,
+so the fidelity oracle can compare what an agent sent against what kitty actually
+put on the wire.  This is the format the Codex path speaks, both as an inbound
+shape and as the upstream shape of the ChatGPT-subscription provider.
+
+**Written against the published schema, never against kitty's output.**  Every
+key, item type and union member below was read from ``openai/openai-openapi``,
+``openapi.yaml``, ``info.version`` :data:`SCHEMA_VERSION` — not from a kitty
+request and not from memory.  §3.3.1's independent-oracle rule: a reader
+validated against kitty's output inherits kitty's bugs, and the whole of I1 then
+proves only that kitty agrees with itself.
+
+**It imports nothing from ``src/kitty``, and must not.**
+``test_reader_responses.py`` asserts the absence structurally.
+
+**What totality means here, and where it stops.**  Every one of the 31 published
+top-level keys is classified into the envelope, the conversation, or the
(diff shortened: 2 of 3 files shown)


shelpuk:
Thanks — two of the three taken, pushed in `32b7ace`. The third I'm declining, with reasoning below.

**1. `additional_tools` had no specific test — agreed, added.**

It was covered only by the parametrised sweep over the user-side opaque items, which asserts the role but not *why*. That is the weakest kind of coverage for this particular entry, because it is the one item carrying a `role` of its own and its only published value is `developer` — which everywhere else in this reader lifts into `conversation.system`.

A future reader seeing `role: "developer"` would reasonably try to lift it, and because paths are index-based, would then move every later turn index. `test_additional_tools_lands_in_a_user_turn_despite_declaring_developer` now records why it cannot: `Conversation.system` is enforced `Text`-only and an `Opaque` has no home in it.

**2. "Nine forms" vs 20 parametrize cases — agreed, clarified.**

Both numbers were right and the docstring said only one of them. Nine *forms* are the members of `ToolChoiceParam.oneOf`; twenty *values* because three of those forms carry closed enumerations of their own (`ToolChoiceOptions` three strings, `ToolChoiceAllowed` two modes, `ToolChoiceTypes` eight built-in types).

The docstring now also says why every value is exercised rather than one representative per form: the normalisation differs *inside* those enumerations — `required` becomes `any` while `auto` does not — so a per-form sample would miss the case that actually varies.

**3. Line number beside the `_cc_to_responses` citation — declining.**

`TEST_SUITE.md` rules this out in its own opening, before §1:

> **On citations.** Symbols are named, not line-numbered. `server.py` is 6,463 lines and line numbers rot on the next edit; `grep -n` or a symbol lookup recovers them, and a stale number is worse than none.

The citation I added is already in the document's house form — `openai_subscription.py:OpenAISubscriptionAdapter._cc_to_responses`, a file-and-symbol pair — which matches how every register row in §3.2.2 cites its site. Adding a line number would make this one entry inconsistent with the ~40 around it and would go stale on the next edit to that file, which is exactly the failure the policy anticipates.

Happy to revisit if the policy itself is up for discussion, but that would be a change to the document's convention rather than to this line.

---

Post-change state: **177 tests** in the new file, 546 across `tests/harness/` plus layer markers, all passing. `ruff check` / `ruff format --check` clean, `mypy src` clean, `lint-imports` 4/4.

shelpuk:
Taken, with one correction to the framing. Pushed in `74c2db9`.

**You were right that the assertion is self-referential.** `test_the_closed_set_matches_the_published_schema_counts` compared `len(r.PUBLISHED_ITEM_TYPES)` against the literal `31` — both in this repository. The **name and docstring were the real defect**: the docstring claimed that "if OpenAI adds an item type, this fails here", which is simply false. The suite has no network and vendors no copy of `openapi.yaml`, so nothing in it can notice OpenAI changing anything. Renamed to `test_the_closed_sets_are_internally_consistent_and_sized_as_recorded`, and the docstring now says what it actually guards.

**Where I'd correct the finding: drift is not uncaught in either direction.**

- *A maintainer edits this module* — a type moved between the role tables, or dropped. Caught by this test, which is the direction a person can cause.
- *OpenAI adds an item type.* Caught at runtime, and loudly: a type outside the closed set residualises, and a non-empty residual **fails the run** (§3.3.1). The first request carrying a new type fails at the point it matters.

That second path is the designed behaviour rather than a gap — §3.3.1's stated intent is that a new wire field must *force a deliberate decision*: "map it, or declare it ignored with a reason." A test that silently widened the tables to match a new spec revision would defeat that. `SCHEMA_VERSION` records which revision the tables were derived from so the decision has a starting point.

The genuinely uncovered case is narrower than the finding suggests: OpenAI adds a type **and** no traffic ever carries it **and** we believe the table is complete. The harm there is a stale comment, not a missed mutation.

**Why not vendor a schema extract to close it properly.** I considered it — there is precedent in `tests/data/`. Two reasons against: a checked-in extract is a second transcription of the same source, so it proves the two copies agree rather than that either matches OpenAI; and it would put a synchronisation burden on all six readers to cover a case that already fails loudly. "Residual empty across the whole corpus" belongs to **T-D8**, which is where cross-reader coverage of this kind is specified.

**One thing I added beyond the finding.** The old assertion did not check that the modelled four and the opaque twenty-seven are *disjoint*. If a type sat in both, dispatch order would silently decide which behaviour won. Now asserted as a partition summing to 31.

---

Post-change: **177 tests**, 546 across `tests/harness/` plus layer markers, all passing. `ruff check` / `ruff format --check` clean, `mypy src` clean, `lint-imports` 4/4.

PR #75, opened 2026-09-11
Read an Anthropic Messages request without asking kitty how (KBR-33 / T-A1)

shelpuk:

Kitty Bridge sits on the wire between a coding agent and whichever LLM provider the user chose. The product promise is that it relays the conversation **faithfully** — it may swap the model name and a few control fields, and nothing else. Today nothing checks that promise.

The check being built is a **fidelity oracle**: read what the agent sent, read what actually went upstream, compare. The two are written in different vendors' formats, so the bytes cannot be diffed directly. Each format therefore needs a small **reader** that translates a request into one common shape; then the comparison is like-for-like.

This PR is one of six such readers — the one for Anthropic's Messages format, which is what Claude Code speaks. With the Chat Completions reader it unblocks the oracle itself ([KBR-51](https://shelpuk.atlassian.net/browse/KBR-51)).

The reader is deliberately written from **Anthropic's published documentation** and imports nothing from kitty. A reader built from kitty's own code would inherit kitty's bugs, and the test would only prove that kitty agrees with itself.

**Two things this work surfaced that matter beyond the test suite** — both filed, neither fixed here:

- **[KBR-167](https://shelpuk.atlassian.net/browse/KBR-167) — blocks the oracle.** Claude Code marks parts of nearly every request as cacheable; that is what stops the user paying full price for the same long prompt every turn. kitty's translator currently **strips those markers**, and the common vocabulary has no field to record them in, so the oracle cannot see the loss. A `grep` while filing it widened the picture: on a translated route kitty re-emits only ten fields, discarding the user's stop sequences, their tool-choice instruction, sampling controls and billing attribution. Only one of those discards is registered as approved. Whether each is a missing register entry or a real defect is exactly what the oracle exists to decide.
- **[KBR-168](https://shelpuk.atlassian.net/browse/KBR-168) — corrected and recommended for closure.** I filed it on a wrong premise and have said so in the ticket rather than deleting it.

---

## What it does

`tests/harness/reader_anthropic_messages.py` projects a captured `POST /v1/messages` into `contract.Request` — envelope, conversation, residual, consumed. It reads **both ends** of the comparison: the inbound body and the captured upstream one, wherever an adapter speaks Messages.

Every top-level field of the published schema is classified, plus `context_management` and `mcp_servers` (real request keys the reference table omits because it omits beta fields). Content blocks: `text`, `image`, `thinking`, `tool_use`, `tool_result` modelled; every other block type kept as a fingerprinted placeholder rather than dropped.

**Totality is the load-bearing property.** Every key at every depth is either mapped and declared, or recorded under its own path — and an unrecorded field fails the run. `verify_total` only sees top-level keys, so this reader's own tests are what close the gap beneath them.

## Four decisions worth their reasons

- **`effort` is recognised although the API reference does not list it.** Claude Code sends it and `MessagesTranslator` reads it off the inbound body, so rejecting it would fail the run on every real request. An earlier draft had this backwards — believed kitty *invented* the key — and generalised a rule for six later readers from that mistake. §7.4.1 now states two evidence rules and says plainly which one has no current example.

- **A wrongly-typed field is recorded, never coerced and never fatal.** `dict(["ab","cd"])` **fabricates** `{"a":"b","c":"d"}` — a tool schema the agent never sent. Aborting instead would blind the oracle to everything else in a request it could otherwise still check. Structural failures do still raise, because there is no partial projection to salvage.

- **`ValueError` is deliberately never caught.** From inside a reader it means the reader mis-routed a field, which must not be disguised as a malformed body — that is the one diagnosis T-D1 depends on.

- **Part order is preserved; nothing is re-sorted.** See the reconciliation below.

## Reconciled with T-A3, which was written in parallel

**T-A3** ([KBR-35](https://shelpuk.atlassian.net/browse/KBR-35)) — the OpenAI Responses reader — was being written in another session and had settled three shared questions, two differently from my draft. **Git would not have caught it:** both changes edit `TEST_SUITE.md` in different places and merge cleanly *while contradicting each other*. I found it by reading the sibling worktree's commits before opening this PR.

| Question | Kept |
|---|---|
| Module layout | **T-A3's** flat `reader_<format>.py` — §3.3.1 already says harness modules sit "beside" the contract |
| The merge rule's third clause | **T-A3's**, and it is a correctness fix, not a coin flip |
| A wrongly-typed leaf | **T-A3's** — record it; both of us had already rejected coercion independently |

The ordering one was a real bug in my draft. `tool_result → user(text) → tool_result` must project as `[ToolResult, Text, ToolResult]`. My re-sort gave `[ToolResult, ToolResult, Text]` — hoisting a result ahead of text the agent sent **before** it, moving history the bridge did not move. Because paths are index-based, that invented delta lands on every later part. There is now a test named for exactly that case.

## Design

Adds **`TEST_SUITE.md` §7.4.1**, which fixes the decisions every later reader inherits: module layout, fail-closed at depth, residual key spelling, the placeholder fingerprint recipe and its cross-format cost, wrongly-typed leaves, tool-result strings, the merge pipeline's ordering, the evidence rule for unpublished keys, and the third outcome §3.3.1 promises but the contract does not implement.

## Evidence

- **103 tests**, every falsification case paired with a control — without the control, a reader that rejected everything would satisfy each positive case.
- **20 hand-applied mutations, all killed, no survivors**: re-sorting after the merge; dropping the merge; every wrong spelling of the fingerprint recipe (`ensure_ascii`, `separators`, `sort_keys`); both coercions; the dropped rename; a dropped `consumed` entry; recording disabled; the media type folded into the image digest; a residualised thinking signature; a placeholder without a fingerprint; cache markers inside it; by-name instead of indexed keys; a JSON-shaped tool result parsed as structured; the catch-all narrowed; `is_error` hard-coded; `tool_choice.name` consumed unconditionally; raising with an empty message.
- **Honest note on process:** the first tranche of tests was written red-first, but the reader was then implemented in one pass, so tranches two and three went green on their first run and were never observed failing. The mutation set above is the fail-first evidence in their place, and it is stronger.
- Full gate green: `4385 passed, 2 skipped`. `ruff`, `lint-imports`, `mypy src/kitty` all clean.
- Layer: inherits `l1` by path; no hand-written marker.

Review history: two `system-design-reviewer` passes (the first caught the `effort` premise error) and one `code-reviewer` pass that returned CHANGES REQUESTED on three defects — a silently dropped field, a fabricated tool schema, and an unguarded fingerprint recipe. All fixed; test count 94 → 103.

https://claude.ai/code/session_01JDdNzJzvNbZzQFzkAQM8Cm

[KBR-51]: https://shelpuk.atlassian.net/browse/KBR-51?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[KBR-167]: https://shelpuk.atlassian.net/browse/KBR-167?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


shelpuk:
## Addressed all three review findings — one was a real defect

Each verified by reproduction before acting, and each now guarded by a test and a mutation.

### 1. `is_error=bool(...)` coerced a wrongly-typed leaf — **fixed, and it was the sharpest of the three**

This one was a genuine defect, not a style point, and it contradicted the rule this very PR adds to §7.4.1.

```python
{"type": "tool_result", "tool_use_id": "t", "content": "x", "is_error": "false"}
# before: is_error is True, residual {}, verify_total passes
```

`bool("false")` is `True`. A body saying **the tool did not fail** projected as a tool that **did** — silently, with an empty residual and a passing totality check. On the response-fidelity path that is a fabricated failure the agent never reported.

It qualifies as a *wrongly-typed leaf* rather than a structural failure by §7.4.1's own test: the grammar has an absent value to fall back to (`False`, the field's default), exactly as `schema` has `None` and `arguments` has `{}`. So it residualises at `<block>.is_error` and carries `False`.

New test `test_a_wrongly_typed_error_flag_residualises_rather_than_being_coerced`, and a new mutation (`coerce a wrongly-typed is_error instead of residualising it`) that the suite kills.

### 2. `IndexError` unreachable in the catch-all — **removed**

Confirmed: the only positional index in the module is `merged[-1]`, guarded by `if merged and ...`. Nothing else indexes a sequence by position — every structural access is dict-keyed or `enumerate`-driven. So the clause was error handling for an impossible case, which `CLAUDE.md` §3 forbids.

Removed, and `read_request`'s docstring now records *why* it is absent, so it does not get added back for symmetry with the other three.

### 3. No test for `cache_control` on a `tool_use` block — **added**

Correct, and worth more than a coverage tick: Claude Code sets a cache breakpoint on tool blocks as well as text ones, so this is the same [KBR-167](https://shelpuk.atlassian.net/browse/KBR-167) blind spot one block type further along. `test_cache_control_on_a_tool_use_block_residualises` now pins it, with its own `ResidualFieldsError` assertion.

---

### Evidence after the fixes

| | |
|---|---|
| Tests | 106 → **108** |
| Hand-applied mutations | **21/21 killed, no survivors** |
| Injection probe | 938 corrupted bodies — **0 non-`UnreadableBodyError` escapes, 0 type-contract violations** |
| `ruff`, `lint-imports`, `mypy src/kitty` | clean |

Pushed as `0bb2f24`.

*Note on the two carryover threads the review flagged as unanswered: they were findings 1 and 2 above. They are now fixed in code rather than argued in a reply.*

[KBR-167]: https://shelpuk.atlassian.net/browse/KBR-167?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

shelpuk:
## Design review found the rule was failing open in eight more places — fixed

A `system-design-reviewer` pass checked §7.4.1's wrongly-typed-leaf rule **against the code** rather than against the three places a bug had been found, and it was right to.

### The finding

§7.4.1 and R7.3 both state the rule generally. The code applied it to three leaves. The contract validates only `Turn.role`, `Conversation.sampling` and `extra["tool_choice"]`, so **every other optional leaf carried whatever the wire sent** — silently, empty residual, passing `verify_total`:

```
thinking.signature: {"a":1}    →  Thinking(signature={'a': 1})     residual {}
tools[0].description: {"a":1}  →  ToolDecl(description={'a': 1})   residual {}
model: {"a":1}                 →  Envelope(model={'a': 1})         residual {}
stream: "yes"                  →  Envelope(stream='yes')           residual {}
```

…plus `tool_use.id`, `tool_result.tool_use_id`, `image.source.media_type` and the image `ref`. Eight in total, all reproduced before acting.

`ToolDecl.description` is **§7.4.1's own worked example** of why a wrong value must not be carried; `signature` and `tool_use_id` are M8's and M7's fields.

All eleven eligible leaves now go through one helper, `_typed_leaf`, so the next field added inherits the rule instead of needing to be remembered. §7.4.1 now says the rule is general and says to apply it that way, for the same reason.

### Why my own probe missed it

The probe asserted "no projected field violates its declared type" — but checked five fields. That is the same species of error as the finding itself: a general claim, a partial implementation. It now checks **every** field the contract declares across all seven part variants plus `ToolDecl` and `Envelope`, and reports **0 violations over 938 injections**.

### Documentation drift, also cleared

The same review catalogued everything left stale by the T-A3 reconciliation, and it was more than I had noticed: six `REQUIREMENTS.md` references still specified the abandoned `projections/` package; §7 item 6 still described the re-sort that R4.8, §11 and §7.4.1 all reject — and §7 is the section a reader consults to learn what §7.4.1 contains, so it pointed them at the defect; four counts were stale; §5 listed two mutations that can no longer be applied; and R7's two worked examples for the catch-all are both handled by named guards now, so the live one replaces them.

### Raised on the sibling ticket rather than resolved by merge order

The review also found that **T-A3 ([KBR-35](https://shelpuk.atlassian.net/browse/KBR-35)) diverges from §7.4.1 in two places** — the structural-vs-leaf boundary, and the unpinned `Opaque` digest recipe. Both branches touch `TEST_SUITE.md` in different places and **will merge cleanly while disagreeing**, which is how the first three divergences between these two tasks nearly landed.

Raised on KBR-35 with the reasoning and an explicit offer to change §7.4.1 instead if T-A3's answer is the better one. Not something to settle by whoever merges second.

---

| | |
|---|---|
| Tests | 108 → **117** |
| Mutations | **22/22 killed** — including the new "drop the `_typed_leaf` guard so optional leaves fail open" |
| Injection probe | 938 injections, **0 escapes, 0 type-contract violations** (now checking every declared field) |
| `ruff` · `lint-imports` · `mypy src/kitty` | clean |

Pushed as `3c4b945`.

[KBR-35]: https://shelpuk.atlassian.net/browse/KBR-35?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

shelpuk:
## `file_id` now tested — the guard existed, the case did not

Correct, and worth taking rather than waving off: `_typed_leaf` guards the image `file_id`, but the parametrised leaf test covered eight of the nine. Deleting that guard — or typo'ing its key — left the suite green.

That is the same shape as the gap which produced the guard in the first place: **a rule applied in code with no case that fails when it is removed.** Three findings across these rounds have now been that shape, which is a fair signal that "is every instance of this rule pinned by a case that dies without it?" belongs in the checklist for the four readers still to come.

Verified rather than assumed — reverting `file_id` to an unguarded `source.get("file_id")`:

```
FAILED test_every_optional_leaf_residualises_when_the_wire_type_is_wrong[...source.file_id]
1 failed, 117 passed
```

Reader restored, 118 tests pass. Pushed as `c8eb708`.

PR #76, opened 2026-09-12
T-W8 (KBR-31): bridge fixture core and the transport extension interface

shelpuk:
Closes [KBR-31](https://shelpuk.atlassian.net/browse/KBR-31) — plan task **T-W8**, Milestone 0.

Kitty Bridge sits on the wire between a coding agent and an LLM provider. Three of its promises are commercial, not cosmetic: the user's message reaches the provider unchanged except where we say otherwise, the bridge is indistinguishable from the agent talking to the provider directly, and nothing escapes the configured network path. **Every test that proves any of those needs a real bridge running in front of a fake provider that records exactly what was sent.**

Today no such thing exists. Of the 62 test files covering the bridge, **38 build their own** — their own fake provider, their own start/stop plumbing, their own request helper. Around **fifteen** planned pieces of work need one shared version, and three more need to plug in *different* provider transports (the ChatGPT subscription path, Amazon Bedrock, Ollama Cloud). This ships that shared version, plus the socket those three plug into without editing shared code.

**Why it took four review rounds rather than one.** A test harness of this kind fails *silently*. Measured on this branch: a bridge pointed at the **wrong live provider** returns a clean **200** — the test passes, and the recording is simply empty. Every later check built on it ("did the user's message survive?", "did anything leak?") would then be measuring nothing and passing for free. You would have a green suite proving nothing about the product's core promises. So the fixture ships with a check that a request genuinely landed in *its* recorder carrying a fingerprint unique to that request, plus four deliberately broken versions it must catch.

**What the reviews caught that would have shipped broken:**

| Found | Consequence if shipped |
|---|---|
| Pointing an adapter at the test provider by replacing its whole URL | Works for Azure, silently destroys **Vertex's routing — where the billed Google Cloud account lives**. The obvious test would have passed |
| The "any new transport inherits the check automatically" promise | Was false for **all three** transports it was written for — each would have crashed on the day it registered |
| One of the four checks | Could be deleted with nothing going red. Removed; the project's own rules forbid an assertion nothing can kill |
| The shared `sample_profile_dict` test fixture | Cannot build the object it describes (wrong ID version) and nothing reads it — filed as **KBR-175**, left for you to decide: delete or repair |

No product code changed. This is test infrastructure plus the design document that was missing for it.

## What is in the diff

- **`tests/harness/bridge.py`** — the fixture, the transport extension interface and registry, the profile/backend factory, the balancing determinism seam, the inbound-request vocabulary, and the conformance check.
- **`tests/harness/test_bridge.py`** — 66 cases covering the core.
- **`tests/harness/test_bridge_falsification.py`** — the four deliberate defects (plan §1.4).
- **`tests/harness/recorder.py`** — `format_for_path` extracted from `_format_for`; the fixture must judge captured paths at teardown, and the original *appends to* the list it would be judging.
- **`.system_design/TEST_SUITE.md` §7.5** — new. §7 had a section for every shared harness and none for this one.

## Evidence

| Check | Result |
|---|---|
| Full suite | **4361 passed, 2 skipped** (19m41s) |
| `ruff`, `mypy src/kitty`, `lint-imports` | clean · clean · 4 contracts kept |
| Layer guards | 68 passed |
| New modules' runtime | **1.6 s** combined, registered in §8.2 with the measurement |
| Falsification | Each conformance assertion deleted in turn; **exactly** its own case went red |
| Assertion liveness | All **80** assertions in the new tests negated one at a time — **0 survivors** |

## Decisions worth a reviewer's eye

- **`redirected()` re-hosts, never replaces** — scheme and authority only, path and query verbatim. §3.3.5 already specifies that substitution from the comparison side, so both sides are now one rule.
- **Two vocabularies, not one.** `WireFormat` is what a recorder serves; `InboundProtocol` is the route an agent posts to. Translating between them is what the bridge *is*, and two of the six formats have no inbound route at all.
- **No `host=` seam, no pytest plugin.** Neither had a named consumer; the plugin measured 0.72 s charged to every pytest run in the repo.
- **`l1`, not `l3`** — §8.2's rule: a test may not move to `l3` before the Subsystem job exists, or it runs in no job at all.

Requirements, acceptance criteria and all 40 review findings with their dispositions are in `.requirements/20260911T233557Z_bridge_fixture_core/REQUIREMENTS.md` (gitignored, local).

https://claude.ai/code/session_016GTtJS1VSesqdAsp66n6fz

[KBR-31]: https://shelpuk.atlassian.net/browse/KBR-31?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.system_design/TEST_SUITE.md
+++ b/.system_design/TEST_SUITE.md
@@ -2211,7 +2211,7 @@ transport (below).
 
 | Piece | What it is |
 |---|---|
-| `UpstreamTransport` | The **extension interface**: `name`, `format`, `start`, `stop`, `bind`, `captures`, `connections`, `assert_teardown_clean` |
+| `UpstreamTransport` | The **extension interface**: `name`, `format`, `start`, `stop`, `bind`, `captures`, `connections`, `assert_teardown_clean`. **Open question for T-B1:** `format` is typed `WireFormat`, which §3.3.1 closes at six — and the `openai_subscription` **OAuth token leg** §7.2 assigns it is none of them. T-B1 decides whether that leg is a separate transport, a recorder outside this interface, or a seventh format; named here so the decision is deliberate |
 | The registry | `register_transport` / `transport(name, fmt, *, responder)` / `registered_transports` |
 | `AiohttpTransport` | The **only** transport registered here, over §7.2's `RecordingUpstream` |
 | `redirected(adapter, origin, provider_config)` | Re-hosts any **default-transport** adapter onto a recorder, keeping path and query |
@@ -2219,7 +2219,9 @@ transport (below).
 | `pin_backend_order` | The determinism seam for balancing mode |
 | `InboundProtocol`, `inbound_path`, `minimal_inbound_body` | The agent-facing side: route and smallest body per inbound protocol |
 | `BridgeFixture` | An async context manager: a real `BridgeServer`, single-backend or balancing, against a started transport |
-| `assert_transport_reaches_its_recorder` | The **integration conformance check**, run over every registered transport by a meta-test |
+| `assert_transport_reaches_its_recorder(transport, *, protocol=None)` | The **integration conformance check**, run over every registered transport by a meta-test. Takes a transport **instance**, not a registry name — so the falsification defects need never be registered, and an author whose format has no matching inbound route names one |
+| `marker()`, `protocol_for()` | The per-request marker, and the inbound route matching an upstream format |
+| `MisdeclaredFormatError`, `TransportTimeout` | What a teardown check and a timed-out request raise |
 
 **Naming.** One list of captured requests, spelled `captures` on a transport and `requests` on
 `RecordingUpstream` (§7.2's own name, unchanged). `CapturedRequest` is the element type.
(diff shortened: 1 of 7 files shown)


shelpuk:
Addressed the review. Two findings were real defects; I verified both before fixing, and falsified both fixes by removing the guard and confirming the new test goes red.

**1. `BridgeFixture.start()` did not release the transport when the *transport* failed.** Correct, and my earlier fix covered only half of it: `RecordingUpstream.start` assigns its runner, awaits `setup()`, and only then binds the site — so a bind failure leaves a runner still needing `cleanup()`, and it happens before there is any bridge for the second guard to unwind. Two guards now, because one genuinely cannot cover both. New test `test_a_transport_that_fails_to_start_is_released_too`; removing the guard makes it fail.

**2. The structural guard could not see a relative import.** Confirmed by probe — `from .epic_b import X` yields `node.module == "epic_b"` with `node.level == 1`, so it matches no `harness.` prefix, and this package has an `__init__.py` so the form is available. The guard now rejects any relative import outright rather than trying to resolve one. Falsified by making the core's own recorder import relative.

The four style and documentation suggestions are all taken: `socket` and `CustomOpenAIAdapter` hoisted to module level, `recorder.py`'s module docstring now names `format_for_path`, and `assert_teardown_clean` carries the `check_*` cross-reference where an author adding a sibling will actually read it — the design doc alone was the weaker place for it.

Gate after the change: 400 harness tests, `ruff`, `mypy src/kitty`, `lint-imports` (4 contracts kept) and both layer guards clean. Full suite was 4361 passed / 2 skipped before these six changes and is re-running in CI.

PR #86, opened 2026-09-12
Stop a model's own prose deciding why a run failed (KBR-166)

shelpuk:
Closes [KBR-166](https://shelpuk.atlassian.net/browse/KBR-166).

Every pull request here is reviewed automatically by an AI reviewer. When that review fails, a small classifier reads the failure record and decides two things: **who has to fix it** — "top up the account" versus "fix the workflow" — and **whether to automatically try again**.

It decided by searching the failure record for tell-tale words. The problem: part of that record is **text the AI reviewer wrote itself**. Four of the search terms were ordinary English words and bare numbers — `quota`, `billing`, `401`, `403`. So a reviewer that merely *wrote the word "quota"* in its own answer changed the verdict on why the run failed.

**That costs real money.** One failure we have on record burned **$1.79 across 24 turns** before dying. Misclassified, it is retried automatically — so we pay the $1.79 a second time for a run that cannot succeed. On two of the four words the operator was also handed contradictory instructions: a headline saying the workflow is broken, and a paragraph underneath telling them to top up a balance that was never spent.

**What changes for the product:** failures now name the right cause, money is not spent twice, and the "top up the balance" instruction appears exactly when it is true. As a side effect the classifier recognises **three provider errors it could not see before**, including Anthropic's documented "your credit balance is too low".

---

## The defect, measured on `main`

`classify` searches `_outcome_text`, which on a record with no structured-output marker includes `result` — the model's own answer. Four patterns it searched were a bare word or a bare number. Measured on the observed `BILLED_INVALID_REQUEST` record with one prose word changed:

| prose word | status | retried | top-up advice |
| --- | --- | --- | --- |
| *(none)* | `fatal` ✅ | no ✅ | no ✅ |
| *"timeout"* | `fatal` ✅ | no ✅ | no ✅ |
| **"quota"** | **`exhausted` ❌** | **yes ❌** | **yes ❌** |
| **"billing"** | **`exhausted` ❌** | **yes ❌** | **yes ❌** |
| **"401"** | **`exhausted` ❌** | **yes ❌** | no ✅ |
| **"403"** | **`exhausted` ❌** | **yes ❌** | no ✅ |

This supersedes both the ticket's table and its correction comment, which were measured before KBR-145 landed.

## Not by anchoring — three attempts are why

The obvious fix is to anchor the four patterns on the vocabulary they were written for. **That was tried three times and each version lost a real body:**

| attempt | what it broke |
| --- | --- |
| named vendor vocabulary only | a CLI 401 whose message is merely `Incorrect API key provided` → `fatal` |
| + structural `\b40[13]\b\s*\{` | the nested carrier `error.message = "HTTP 401 Unauthorized"` → `fatal` |
| + anchored quota word | 4 of 10 real quota wordings → `fatal` when the body carries the generic code |

Every one is **KBR-145's filed defect one tier across**: once a provider-naming tier stops answering, `FATAL_UNLESS_PROVIDER_NAMED_PATTERNS` claims the record and an operator with a dead key or a spent balance is sent to debug a correct workflow. It reproduces, three more times, what KBR-172 recorded for the eight bare status codes: *no anchor matches every real body while leaking no prose.*

## The mechanism: not what the text says, but who wrote it

`_outcome_text` already answers that. KBR-172 built the split — `provider_parts` versus `model_parts` — and applies it **conditionally**. This change applies the same split **unconditionally**, to the weak patterns only:

* `QUOTA_WORD_PATTERNS` = (`quota`, `\bbilling\b`)
* `CREDENTIAL_STATUS_PATTERNS` = (`\b40[13]\b`)

both searched over a new `_provider_outcome_text` and consulted immediately after their anchored counterparts — the tier position the moved patterns already held. **Nothing is deleted and no pattern is rewritten.**

Three anchored patterns are added, each closing a body that names a cause but carries no number (reachable because of KBR-182): `billing_error`, `\bauthentication_error\b`, `\bpermission_error\b`, all from Anthropic's published error reference.

### `_provider_outcome_text` has its own fallback contract

Inheriting `_outcome_text`'s is wrong in **both** directions, and this is the subtle part:

* Searching an unparseable record whole is justified there as *"a CLI-level failure message with no tool results in it"* — an argument about `message`/`content`, not `result`. Measured: a pretty-printed array with one line appended fails `_parse_events`, and the whole text including `result` then reached the status tier.
* Returning `""` is equally wrong: the CLI-prefix carrier is **also** unparseable and is the only path by which a 401 with no credential vocabulary is recognised at all.

So the two cases are split by the fact that distinguishes them — raw CLI output has no `result` **key**, a broken record does. The `\s*:` is load-bearing: every result event carries `"type": "result"` as a *value*.

## Verification

* **664 tests pass** (648 → 664), under both `pytest` and CI's own `python …` invocation.
* **13 of 13 falsification mutants killed.** Every part of the fix was broken in turn and every break was caught. Two mutants survived the first sweep — the missing `\s*:` coverage and an unpinned `\bpermission_error\b` — both found by code review and now pinned.
* All ten real quota wordings classify **in the hostile carrier** with their advice; none regresses.
* Six credential carriers classify, two of which did not before.
* Verdict-and-advice agreement swept over 138 generated records: 0 violations.

`ruff` deliberately excludes `.github/` (`exclude = [".github"]`) because that tree is carried byte-for-byte from another repository; no reformatting was done there. The 648 → 664 bump in three documents is required by the existing `test_every_document_quotes_the_real_count` guard.

## Filed while measuring, deliberately not fixed here

Per the standing policy in `.system_design/TEST_SUITE.md`:

* **[KBR-181](https://shelpuk.atlassian.net/browse/KBR-181)** — five more quota patterns leak the same way, two of them matching **this test file's own live line numbers** (it is 16,900+ lines, so `test_review_scripts.py:1308` is a citation a reviewer really would write). Grandfathered by name in a frozen guard so a sixth cannot be added quietly.
* **[KBR-182](https://shelpuk.atlassian.net/browse/KBR-182)** — `api_error_status` is listed as an outcome field and never read, because JSON sends status codes as numbers and `_strings_in` collects strings only.

## Residual exposure, stated rather than hidden

1. A **provider** writing `quota`/`401` as prose in its own error message still reads as that signal. This module can say who wrote a field; it cannot say whether a provider meant its own word. A product-owner decision, taken on measured costs.
2. A review that quotes a **verbatim provider body** still matches it — irreducible and pre-existing, shared by `insufficient balance` and `authentication_failed`.
3. An unparseable record carrying a `result` key gets no provider-scoped match and is `fatal` with no retry. Deliberate: it spends nothing on a record that cannot be read.
4. `EXHAUSTED_PATTERNS` keeps its bare words on purpose — that tier sits *below* the generic one, so a prose word there cannot promote a billed rejection. Recorded in the module so it cannot read as an oversight.

https://claude.ai/code/session_01HdkFBQQRUhxJaL2Sj1Ep4t

[KBR-166]: https://shelpuk.atlassian.net/browse/KBR-166?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[KBR-181]: https://shelpuk.atlassian.net/browse/KBR-181?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/review/README.md
+++ b/.github/review/README.md
@@ -205,7 +205,7 @@ step 2 above.
 ## Changing the review system
 
 `.github/workflows/ci.yml` runs `tests/test_review_scripts.py` on every pull
-request — 648 tests over the selector, the classifier, the notices, the redactor,
+request — 664 tests over the selector, the classifier, the notices, the redactor,
 the schema and the workflow's own wiring, plus the workflow parser, the
 runner-ceiling table and the `ci-required` aggregation. Run them locally the same
 way:
--- a/.github/review/scripts/interpret_claude_result.py
+++ b/.github/review/scripts/interpret_claude_result.py
@@ -165,9 +165,45 @@ QUOTA_PATTERNS = (
     # The wording alone catches the real 402 -- its body carries
     # "Insufficient credits" -- so the code bought nothing and cost that.
     r"insufficient credits",
-    # Generic quota and billing.
-    r"quota",
+    # upstream: Anthropic's documented 402 type. `\bbilling\b` could never match it --
+    # `_` is a word character -- so the one status that means "your balance is the
+    # problem" reached nothing in this set. See `platform.claude.com/docs/en/api/errors`.
+    r"billing_error",
     r"exceeded your current",
+)
+
+# 🔴 **KBR-166. The same words, searched ONLY over what the PROVIDER wrote.**
+#
+# `quota` and `\bbilling\b` lived in the set above until this ticket, where they were
+# searched over a haystack that includes `result` -- the model's own answer. So a
+# reviewer that merely WROTE "quota" turned a $1.79 billed rejection into `exhausted`,
+# and `retry_verdict`'s exhausted branch then spent the $1.79 again.
+#
+# ⚠️ **They are MOVED, not anchored, and three attempts to anchor them are the reason.**
+# Anchoring was measured three times -- on named vendor vocabulary, on the CLI's
+# `API Error: NNN {` line shape, and on the word beside a spending verb. Every version
+# lost a real body, and each loss landed in `FATAL_UNLESS_PROVIDER_NAMED_PATTERNS`: an
+# operator with a spent balance sent to debug a correct workflow, which is KBR-145's
+# filed defect one tier across. Measured: an anchored quota word missed four of ten real
+# wordings, and all four became `fatal` with no retry because the body also carried the
+# generic error code -- exactly what DeepSeek sends for a spent balance.
+#
+# That reproduces, a third time, what KBR-172 recorded for the eight bare status codes:
+# **no anchor matches every real body while leaking no prose.** The separable question is
+# not what the text says but WHO WROTE IT, and the record already answers it -- so the
+# fix is the condition KBR-172 built, applied unconditionally to the weak patterns only.
+#
+# ⚠️ Residual, stated so it cannot read as an oversight: a PROVIDER that writes "quota"
+# as prose in its own error message is still read as a quota failure. This module can say
+# who wrote a field; it cannot say whether a provider meant its own word.
+#
+# ⚠️ One self-match, recorded rather than discovered later: `classify` returns
+# "provider quota exhausted: ...", which `quota` matches. Not live -- that string goes to
+# `$GITHUB_OUTPUT` and the diagnostic, neither of which re-enters a haystack -- but
+# `OUTCOME_FIELDS` warns about exactly this class, so the next person measuring it should
+# find it written down.
+QUOTA_WORD_PATTERNS = (
+    r"quota",
     r"\bbilling\b",
 )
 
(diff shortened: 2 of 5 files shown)


shelpuk:
## Review finding addressed — `83c0f4e`

**Valid, and fixed.** `_provider_outcome_text`'s docstring carried two `Args:` sections. My own edit caused it: the paragraph recording the first fallback branch's real reach was inserted *above* the existing `Args:` along with a second copy of it.

It is not cosmetic. `sphinx.ext.napoleon` renders the last section and silently drops the earlier one, so the rendered API reference and the source disagree about where the parameter is documented — in a module whose docstrings are the reason anyone can follow the tier order.

Fixed by deleting the duplicate; the surviving block is the one immediately above `Returns:`, in the module's usual order.

**Swept for the same slip rather than fixing only the instance reported.** Both touched files, every function, class and module docstring, across `Args:`, `Returns:` and `Raises:` — no other occurrence.

Suite re-run after the change: **664 tests, OK**, under CI's own `python .github/review/tests/test_review_scripts.py` invocation.

No other findings were raised. The one thing worth noting for a human reviewer is that the review correctly read this as a single-author PR with no earlier rounds — the three rejected anchor designs and the two coverage holes found before it (the load-bearing `\s*:` in the fallback sentinel, and an unpinned `\bpermission_error\b`) were caught before the branch was pushed, so they appear in the description rather than in this thread.

PR #87, opened 2026-09-12
Give the cache breakpoint a home, and register the strip (KBR-167 / T-W2)

shelpuk:

**Claude Code sets a prompt-cache breakpoint on nearly every request.** That is what keeps a long session's token bill down: the stable prefix — system prompt, tool definitions, conversation history — is billed at a cache read of 0.1× the normal input rate instead of full price.

**Two things follow, and this PR addresses the first.**

1. **The test suite could not run against a single real Claude Code request.** The suite's fidelity oracle reads every request into a wire-independent form and fails the run if it meets a field it cannot account for. That form had no home for `cache_control`, so every real body failed. This PR fixes that. The oracle can now be pointed at real traffic.

2. **kitty destroys those breakpoints on every translated route** — including the default Anthropic-to-Anthropic route, because it still detours through Chat Completions. That is a live commercial defect, not a test-suite one. It is **not** fixed here; it is characterised across all four test layers in new epic **[KBR-197](https://shelpuk.atlassian.net/browse/KBR-197)** (tasks KBR-198…KBR-203).

**Why this PR matters beyond unblocking the suite.** KBR-167 offered two ways to stop the run failing, and the choice was the ticket's real content. The rejected option — declaring the field "ignored" — would also have worked, and would have made the strip **permanently invisible to the oracle**. The option taken gives the field a slot and adds register row **M16** claiming the strip, so a destroyed breakpoint is reported as a named, attributable delta. The money stays on the record.

---

## What changed

**The grammar** (`tests/harness/contract.py`) — an optional `cache_control` on the six block types Anthropic permits a breakpoint on: `Text`, `ToolUse`, `ToolResult`, `Image`, `Opaque`, `ToolDecl`. The wire mapping is carried whole rather than flattened to a boolean, because a 1-hour write costs 2× base input against 1.25× for the default, so a silently downgraded TTL has to stay visible.

`Thinking` and `Json` get **no** slot, and the two exclusions have different reasons — conflated, they invite a later "tidy-up":
- `Thinking` — Anthropic does not support a breakpoint there.
- `Json` — it is the part for Converse and Gemini's natively-structured values, and neither format has a `cache_control` concept, so no reader could ever fill the slot.

A breakpoint nested inside a `ToolResult` also residualises: §3.3.1a defines no path form reaching inside one, so M16 could never claim such a delta.

**The path vocabulary** — `system_path` and `part_path` take an optional field name, as `tool_path` already did. M16 names `.cache_control`, never the block, because the coarser anchors would also claim a deleted part and a deleted tool description — two of the design's own five oracle falsification cases.

**The reader** (`reader_anthropic_messages.py`) — maps the field at every site it can appear.

**The register** — row **M16**, trigger `NON_NATIVE_UPSTREAM_WIRE`, `conditional=False`. Unconditional for M2's reason rather than M1's: they share this trigger, and an existing guard forbids two rows disagreeing about whether a shared trigger is conditional. The native-passthrough route *is* a real complement, but it is a property of the route, not of the request, and no corpus entry can choose a provider.

---

## Verification

- `pytest tests/harness` — **767 passed**
- Layer guards — **68 passed**
- Full `-m "l1 or l2"` gate before the merge commit — **4769 passed, 2 skipped**, exit 0 (re-running post-merge)
- `ruff check` clean; `lint-imports` 5 contracts kept, 0 broken
- The ticket's own reported body: **4 residual keys → empty residual**, `verify_total` passes

Reviewed by `system-design-reviewer` (6 findings applied) and `code-reviewer` (**APPROVE**; 4 documentation findings applied).

**Four existing reader tests were inverted, not deleted** — they asserted the old residualising behaviour, which is exactly what a regression would restore. Four register guards moved with the new row count; one of them used `M16` as its synthetic "id that does not exist", which this PR made real, so it moved to `M99`.

---

## Filed, not fixed here

- **[KBR-197](https://shelpuk.atlassian.net/browse/KBR-197)** — epic: prove the bridge does not silently destroy the user's prompt cache. Six tasks, one per layer, with a sourced Anthropic + OpenAI reference dossier. Includes **KBR-203**, a second defect found on the way: kitty's `thinking` rewrite derives `budget_tokens` from `max_tokens`, and Anthropic renders that config into the prompt — so a breakpoint that *survives* can still miss, which bites the native route too.
- **[KBR-205](https://shelpuk.atlassian.net/browse/KBR-205)** — four more Anthropic block extras (`citations`, `transformations`, `caller`, `toolset_name`) still fail the run. Measured, not assumed. They need the design's promised third outcome — a declared-ignored mechanism — rather than more vendor-specific slots, and that is the owner's call.

Closes KBR-167.

[KBR-197]: https://shelpuk.atlassian.net/browse/KBR-197?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


shelpuk:
Heads-up from PR #80, which hit KBR-208 on its own Windows leg: **`c1645d8` fixes the cause but leaves five exemption rows that the fix makes obsolete, and obsolete rows fail the gate by design.**

`tests/exemptions.py` is untouched by this PR. Five of its six rows are keyed on exactly the coarse clock `perf_counter()` removes:

| Row | Issue |
|---|---|
| `recorder-arrival-increases-per-session` | KBR-188 |
| `recorder-arrival-increases-across-requests` | KBR-188 |
| `recorder-falsification-reordering-only-order-check` | KBR-188 |
| `recorder-falsification-miscount-only-connection-check` | KBR-188 |
| `recorder-falsification-probe-passes-real-recorder` | KBR-188 |

(The sixth, `recorder-responder-abort-emits-before-dropping`, is KBR-189 — an abort race, not the clock. It stays.)

Each says the assertion is **false on Windows**. With `arrival` on `perf_counter()` two adjacent requests get distinct timestamps, so those assertions become **true** — and `ratchet` then raises `UnexpectedExemptionPass`: *"The defect is fixed, so delete the row from the registry."* The leg goes red again, in the opposite direction and for a good reason.

### This is not speculation — PR #80 already ran it

On [#80](https://github.com/Shelpuk-AI-Technology-Consulting/kitty-bridge/pull/80) I saw the Windows leg fail with **both** shapes across two runs of the same code, no change in between:

| Run | Outcome |
|---|---|
| 1 | `UnexpectedExemptionPass` on `recorder-arrival-increases-per-session` **and** `recorder-falsification-miscount-only-connection-check` — the clock happened to tick between requests |
| 2 | Neither fired; the raw `check_arrival_increases` failure remained |

So on a 15.6 ms tick that outcome is occasional. On a ~100 ns clock it becomes the **normal** case for all five rows at once.

### Suggestion

Delete those five rows in this PR, for the same reason the clock fix is being carried here: they are one change, and splitting them leaves the Windows leg red either way. KBR-188 is then resolved by this PR rather than left describing a defect that no longer exists.

Worth keeping §8.3’s own framing in mind — the registry is acknowledged debt that should trend to zero, and the unexpected-pass guard exists precisely so a row cannot outlive its cause. This is that guard about to do its job.

PR #91, opened 2026-09-12
Close the transport a custom-transport adapter owns (KBR-190)

shelpuk:
Fixes [KBR-190](https://shelpuk.atlassian.net/browse/KBR-190).

Kitty Bridge sits on the wire between a coding agent and an LLM provider. Most providers ride on the bridge's own network connection, which the bridge closes cleanly when it shuts down. Three providers — **Ollama Cloud**, the **ChatGPT subscription** path, and **AWS Bedrock** — carry their own connection instead, and nobody was closing those.

**What a user would have seen.** Almost nothing, in normal use: the bridge lives as long as the process, so the operating system reclaims the sockets when it exits, and the only symptom is a warning at shutdown. It bites where a bridge is started and stopped **inside** a long-running process — each cycle abandoned a pool of network connections. Our own test suite does exactly that thousands of times per run, which is how it surfaced.

**What changed for the product.** Every provider now answers one instruction — *release your connection* — and the bridge issues it to every provider it owns as part of shutting down, after in-flight requests have finished. One provider misbehaving cannot stop the others, and cannot cost the bridge its own cleanup.

**What this is not.** No change to how requests are served, routed, translated or authenticated. Nothing a user sends or receives moves. This is shutdown hygiene, and its value is that a long-lived bridge stops accumulating dead connections.

**Three follow-ups found while here**, filed and linked rather than folded in: [KBR-210](https://shelpuk.atlassian.net/browse/KBR-210), [KBR-211](https://shelpuk.atlassian.net/browse/KBR-211), [KBR-212](https://shelpuk.atlassian.net/browse/KBR-212).

---

## What was wrong

`BridgeServer.stop_async` promised, in its own docstring, to *"close the HTTP client session"*, and closed the two the bridge builds. The three adapters that set `use_custom_transport` bypass those entirely, and two of them cache a client of their own that **nothing** closed:

| Adapter | Cached client | Closed by, before this change |
|---|---|---|
| `ollama_cloud` | `aiohttp.ClientSession` | nobody |
| `openai_subscription` — serving | `curl_cffi.AsyncSession` | nobody |
| `openai_subscription` — OAuth refresh | `curl_cffi.AsyncSession` | nobody |
| `bedrock` | *none cached* | n/a |

The ticket names one ChatGPT session; there are **two** — the refresh leg has had its own since KBR-161.

## What this does

`ProviderAdapter.aclose()` — a no-op default, overridden by the two adapters that own a client, awaited by `stop_async` for every adapter the bridge holds.

`bedrock` gets **no** override, and that is a decision rather than an omission: `_get_boto3_client` builds a client per request and caches nothing on the instance. `TestItCachesNoTransport` pins that property directly, and the registry sweep in `test_wire_shape_honesty.py` names bedrock as the one exemption — so a fourth custom-transport adapter has to classify itself rather than inherit silence.

### Three things this had to get right

**Order.** `_runner.cleanup()` drains in-flight handlers first, then the bridge's own sessions and state file, then the adapters. A **failed drain deliberately skips** the adapter close: an undrained handler may still hold a live stream, and closing a `curl_cffi` session under one is the shape [curl_cffi #675](https://github.com/lexiforest/curl_cffi/issues/675) reported as a segfault. That issue was filed against 0.13.0 and closed upstream as no longer reproducible, and we pin 0.16.3 — but #845 still tracks #751, so leaking a pool beats crashing the process.

**Containment, in both directions.** One adapter's failure is logged and contained, so the others still close. And the adapters close from a `finally`, so a failure in the bridge's *own* teardown cannot skip them — which would leak exactly what this fixes. `asyncio.CancelledError` is a `BaseException`, so cancellation still propagates rather than being swallowed (verified on the pinned interpreter).

**Detach before closing.** Each override clears its cached attribute **before** awaiting `close()`. The `curl_cffi` builders test for absence only, so a session left in place after a failed close would be handed back for the rest of the process's life — and the containment above would make that silent as well as permanent. This was a blocker raised against my own first implementation.

Teardown reads the backing fields `_provider` / `_backends`, never the `_active_*` properties: those resolve through a request-scoped `ContextVar`, and teardown is not a request. Deduplicated by identity, because balancing mode is constructed with `provider=backends[0][0]`.

## Two recorded decisions are reversed — deliberately, and in writing

Neither is contradicted silently; both docstrings are rewritten in the same change.

1. **T-B1's transport** argued its reach into `adapter._session` belonged in the harness, because *"the product's session is a per-process pool whose lifetime is the process."* False as measured: `get_provider` returns a fresh adapter per call, and all five production call sites build theirs immediately before the `BridgeServer` that receives them. The private reach is gone. The **call** remains, on the public hook, because four tests in that module drive the adapter with no bridge at all and would otherwise leak a session each.
2. **`_curl_session`'s** *"never explicitly closed during normal operation"* note is replaced by the measured basis above.

`.system_design/TEST_SUITE.md` §5.5 gains a **who closes** column covering all five outbound paths, plus the ownership contract. The two KBR-161 passages arguing from *"the process lifetime"* are re-derived — that window is the bridge's now; the separation argument survives intact, only the window shrinks.

## Testing

All L1, per §2.2 — nothing here touches a socket or a clock, and §2.2's own worked example places a test that must construct a `BridgeServer` at L1. New file `tests/bridge/test_adapter_transport_close.py` lands at `l1` by path default, so no layer-registry edit.

**Every new test was watched failing first**, and the load-bearing ones were falsified against a deliberate defect — the repo's standing rule is that a guard no defect can kill is decoration:

| Defect injected | Test that died |
|---|---|
| `stop_async` no longer closes adapters | the harness test, alone in its module |
| teardown reads `_active_provider` | `test_it_reads_the_backing_fields_not_the_request_context`, and nothing else |
| adapters leave the inner `finally` | `test_a_failing_bridge_teardown_still_closes_the_adapters` |
| adapters move to the **outer** `finally` | `test_a_failed_drain_deliberately_skips_the_adapters` |
| `aclose` closes then clears | the three raising-close tests |
| `OllamaCloudAdapter.aclose` deleted | the registry sweep, alone in its file |
| transport releases the port before closing | `test_the_session_closes_before_the_port_is_released` |
| bedrock caches its client | both `TestItCachesNoTransport` cases |
| a client stored but not reused | `test_no_client_is_stored_on_the_instance`, alone |

**Gate: 5191 passed, 2 skipped**, plus `ruff`, `lint-imports` (5 contracts kept) and `mypy` (90 files) clean.

## Review

`system-design-reviewer` twice — six blockers, including one live in code already written. `code-reviewer` ran a **19-mutation sweep** of its own against these tests; both of its findings are applied:

- an assertion-free idempotence test that no defect could *uniquely* kill — dropped, by the rule above;
- a bedrock docstring stating a reason I had not measured — replaced by the one I did, and then measured.

[KBR-190]: https://shelpuk.atlassian.net/browse/KBR-190?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[KBR-210]: https://shelpuk.atlassian.net/browse/KBR-210?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.system_design/TEST_SUITE.md
+++ b/.system_design/TEST_SUITE.md
@@ -1413,7 +1413,7 @@ whether to change it is Q3.
 
 ### 5.5 Per-transport containment
 
-`_session_for` and `should_bypass` govern **only** `BridgeServer`'s own aiohttp sessions. Four
+`_session_for` and `should_bypass` govern **only** `BridgeServer`'s own aiohttp sessions. Five
 other outbound paths exist, and each applies the proxy **unconditionally, without consulting
 `should_bypass`**:
 
@@ -1445,8 +1445,10 @@ custom-transport adapter has to decide about its client as well as its wire shap
 1. **Closing a `curl_cffi` session under a live SSE stream is the one shape that has crashed.**
    `_curl_session`'s docstring recorded that as the reason it was never closed, citing
    lexiforest/curl_cffi **#675**. That issue was filed against **0.13.0** on the *synchronous*
-   `Session` and was **closed 2026-07-18 as no longer reproducible**; this project pins
-   **0.16.3**. More to the point, `stop_async` closes adapters only **after**
+   `Session` and was **closed 2026-07-18 as no longer reproducible**. That was read on
+   **0.16.3**, the version resolved here — `pyproject.toml` declares `curl_cffi>=0.7` with no
+   upper bound, the weakest pin in the repo, so this is measured rather than guaranteed. More
+   to the point, `stop_async` closes adapters only **after**
    `_runner.cleanup()` has drained the in-flight handlers, so the precondition is not met on
    the ordinary path. Upstream **#845** still tracks **#751**, "active stream/session close
    lifecycle", so the risk is reduced, not zero — which is why a **failed drain deliberately
--- a/.system_design/TEST_SUITE.md
+++ b/.system_design/TEST_SUITE.md
@@ -1417,13 +1417,53 @@ whether to change it is Q3.
 other outbound paths exist, and each applies the proxy **unconditionally, without consulting
 `should_bypass`**:
 
-| Path | Client | How the proxy is applied |
-|---|---|---|
-| `openai_subscription` — serving | `curl_cffi.AsyncSession` | `proxies=` + `CURLOPT_NOPROXY` |
-| `openai_subscription` — OAuth **refresh** leg | its own `curl_cffi.AsyncSession` | `proxies=` + `CURLOPT_NOPROXY` |
-| `openai_subscription` — OAuth **login** leg (`kitty.auth.openai_oauth`) | its own `aiohttp.ClientSession` | `aiohttp_session_kwargs()` |
-| `bedrock` | boto3 / botocore | `BotoConfig(proxies=egress.proxies_dict())` |
-| `ollama_cloud` | its own `aiohttp.ClientSession` | `aiohttp_session_kwargs()` |
+| Path | Client | How the proxy is applied | Who closes it |
+|---|---|---|---|
+| `openai_subscription` — serving | `curl_cffi.AsyncSession` | `proxies=` + `CURLOPT_NOPROXY` | `aclose()`, from `stop_async` |
+| `openai_subscription` — OAuth **refresh** leg | its own `curl_cffi.AsyncSession` | `proxies=` + `CURLOPT_NOPROXY` | `aclose()`, from `stop_async` |
+| `openai_subscription` — OAuth **login** leg (`kitty.auth.openai_oauth`) | its own `aiohttp.ClientSession` | `aiohttp_session_kwargs()` | `run_oauth_flow`, in a `finally`, when it built the session itself |
+| `bedrock` | boto3 / botocore | `BotoConfig(proxies=egress.proxies_dict())` | nobody — nothing is cached; `_get_boto3_client` builds one per request |
+| `ollama_cloud` | its own `aiohttp.ClientSession` | `aiohttp_session_kwargs()` | `aclose()`, from `stop_async` |
+
+**The fourth column is new, and it was empty until KBR-190.** `BridgeServer.stop_async`
+closed the two sessions the bridge itself builds and nothing else, so a bridge started and
+stopped **inside a living process** leaked one connection pool per cycle per custom-transport
+adapter. Ordinary use never noticed — the bridge's lifetime is the process's and the OS
+reclaims the sockets — but the suite starts and stops bridges in-process thousands of times,
+and that is where it surfaced, as an `Unclosed client session` per test.
+
+**The bridge owns the adapters it is given.** `ProviderAdapter.aclose()` is the hook: a no-op
+by default, overridden by the adapters that own a client. The ownership is a contract rather
+than an observation about today's call sites — `get_provider()` returns a **fresh instance per
+call** and all five construction sites build their adapters immediately before the
+`BridgeServer` that receives them, so a caller must not hand one adapter to two bridges whose
+lifetimes overlap. `tests/test_wire_shape_honesty.py` sweeps the registry so a fourth
+custom-transport adapter has to decide about its client as well as its wire shape.
+
+**Three points where this could have gone wrong, and what settles each.**
+
+1. **Closing a `curl_cffi` session under a live SSE stream is the one shape that has crashed.**
+   `_curl_session`'s docstring recorded that as the reason it was never closed, citing
+   lexiforest/curl_cffi **#675**. That issue was filed against **0.13.0** on the *synchronous*
+   `Session` and was **closed 2026-07-18 as no longer reproducible**; this project pins
+   **0.16.3**. More to the point, `stop_async` closes adapters only **after**
+   `_runner.cleanup()` has drained the in-flight handlers, so the precondition is not met on
+   the ordinary path. Upstream **#845** still tracks **#751**, "active stream/session close
+   lifecycle", so the risk is reduced, not zero — which is why a **failed drain deliberately
+   skips the adapter close**. Leaking a pool beats crashing the process.
+2. **A failure must not cascade, in either direction.** One adapter's `aclose()` failing is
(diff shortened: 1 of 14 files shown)


shelpuk:
## CI: `ci-required` is red on the Windows leg, and it is not this change

**Every other leg passes** — Linux 3.10, 3.11, 3.12, 3.13, macOS 3.12, plus lint, import layering, types, CodeQL and the review jobs. Only `test / Python 3.12 on windows-latest` fails, with two tests from the **KBR-188** Windows-clock family.

### The evidence that it is pre-existing

`main` at **`11a5c882`** — the exact commit this branch is based on — [already fails its own Windows leg](https://github.com/Shelpuk-AI-Technology-Consulting/kitty-bridge/actions/runs/34699341852), and one of its two failures is character-for-character the same test as one of mine:

| | `main` @ `11a5c882` | this PR |
|---|---|---|
| **Failure 1** | `test_provider_aiohttp.py::TestTheRecorderPassesEveryConformanceCheck::test_per_session_check[check_arrival_increases]` — `got [2244.234, 2244.234]` | **the same test** — `got [1360.75, 1360.75]` |
| **Failure 2** | `test_recorder.py::TestCaptureFidelity::test_arrival_increases_across_requests` — `UnexpectedExemptionPass` | `test_recorder.py::TestTheRecorderPassesEveryConformanceCheck::test_per_session_check[check_arrival_increases]` — `UnexpectedExemptionPass` |

Failure 1 is identical. Failure 2 lands on a different cell each run, which is itself the finding below.

### Two distinct defects, both KBR-188's, both on `main`

**a. One cell was never exempted.** Both `ratchet("recorder-arrival-increases-…")` calls live in `tests/harness/test_recorder.py` (lines 121 and 318). `test_provider_aiohttp.py` runs the *same* conformance check through the provider recorder and carries no exemption, so it fails bare on Windows. A coverage gap in KBR-188's exemption set, not a new break.

**b. A binary exemption cannot hold a nondeterministic defect.** The mechanism is deliberately two-sided — an unexpected pass fails the job, so an exemption cannot outlive the defect it describes. That is right for a deterministic defect. Windows' clock granularity is ~15.6 ms, so whether two adjacent requests share a timestamp is a coin flip per cell per run. On `main`'s run one cell got distinct stamps and failed as an unexpected pass; on this PR's run a *different* cell did. Neither ordering is reachable by either branch's code.

I have added both findings to KBR-188 rather than working around them here — they are that ticket's, and patching the exemption registry from a KBR-190 PR would hide a live defect behind an unrelated change.

### What this change actually touches on Windows

Nothing timing-related. `stop_async` gained an adapter-close step **after** the runner drain, and the two failing tests assert on request *arrival* timestamps recorded during the run, before any teardown. The local full gate on Linux is **5191 passed, 2 skipped**, and the four other platform legs here agree.

Flagging rather than deciding: `ci-required` will stay red until KBR-188 is resolved, so merging this needs either that fix to land first or a deliberate call from a maintainer.

shelpuk:
All three review findings applied in `8ca2508`, each verified against the code first.

**1. The version claim was false — and in the way this PR filed a ticket about.** The reviewer is right: `pyproject.toml:29` declares `curl_cffi>=0.7` with no upper bound, which `tests/test_curl_cffi_transport_contract.py` already calls the weakest pin in the repo. **0.16.3 is what resolved here and what I read the behaviour on, not what is pinned.** That distinction is load-bearing rather than pedantic — the whole safety argument for closing these sessions rests on #675 being fixed, and a future resolution could pick up a version where the lifecycle work still open under **#751** bites. Both the docstring and §5.5 now say *measured, not guaranteed*, name the declared range, and connect it to why a failed drain skips the close entirely.

I had filed [KBR-212](https://shelpuk.atlassian.net/browse/KBR-212) earlier in this same change for a docstring whose conclusion was right and premise wrong. This was the same error, in text I wrote. Good catch.

**2. §5.5's prose contradicted its own table.** "Four other outbound paths exist" above five rows — stale since KBR-161 split the OAuth leg in two, so pre-existing rather than introduced here. Fixed anyway: the table is the thing this PR restructures, and leaving the count wrong would make the new column read as the error.

**3. The two failing-close tests overlapped.** Also correct, and the cause is worth recording: `assert oauth.closed` was added to `test_both_attributes_are_cleared_even_when_a_close_fails` only to satisfy ruff's F841 on an unused binding, and it quietly swallowed the entire claim of `test_the_oauth_leg_still_closes_when_the_serving_leg_fails` — which could then no longer fail on its own. A lint fix that cost a test its independence.

Split so each owns one logical fact, and **falsified separately afterwards** rather than assumed:

| Defect injected | Test that died |
|---|---|
| drop the `finally` in `aclose` | `test_the_oauth_leg_still_closes_when_the_serving_leg_fails`, alone |
| clear the attributes only on success | `test_both_attributes_are_cleared_even_when_a_close_fails`, alone |

Gates after the change: `ruff`, `lint-imports` (5 contracts kept), `mypy` (90 files) clean; `tests/providers` + `tests/harness` + `tests/bridge` + the provider, layer, exemption and design-doc-consuming modules — **1724 passed**, no failures. The Windows leg remains red for [KBR-188](https://shelpuk.atlassian.net/browse/KBR-188), unchanged and unrelated, as set out above.


[KBR-212]: https://shelpuk.atlassian.net/browse/KBR-212?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[KBR-188]: https://shelpuk.atlassian.net/browse/KBR-188?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #96, opened 2026-09-12
Give prompts and menus one answer to "can this process interact?" (KBR-204)

shelpuk:
Fixes [KBR-204](https://shelpuk.atlassian.net/browse/KBR-204). Follow-up filed: [KBR-218](https://shelpuk.atlassian.net/browse/KBR-218).

**What was broken.** kitty's interactive commands (`kitty egress`, `kitty profile`, `kitty auth openai`) are supposed to refuse politely when they are not run in a real terminal — "requires an interactive terminal (TTY)", exit code 2 — so that a script or CI job gets a clear answer. Half of kitty checked this properly (the menus, fixed last round in KBR-187); the other half (the prompts and command entry points) only checked the keyboard side and ignored where the output was going.

**What a user saw.**
- With output piped or redirected, `kitty egress` and `kitty profile` **exited 0 having done nothing**. A script calling them was told "success".
- On Windows, `kitty auth openai` with redirected output could **run the whole browser login and then crash** with a traceback at the first question.

**What changes for users.** Every interactive command now gives the same clear refusal (exit 2) when its output is not a terminal, on Linux, macOS and Windows. The one visible behaviour change is deliberate: a script that ran `kitty egress` or `kitty profile` with piped output and read exit 0 now reads exit 2 — which is the truth, since nothing happened.

**What is not fixed yet.** On Windows, sending output to the special `NUL` device still fools the check (a known Windows quirk, [bpo-28654](https://bugs.python.org/issue28654)). That needs a different technique, so it is filed separately as KBR-218.

## What changed

- **One rule instead of two.** `kitty.tui.prompts.can_interact()` is now the only answer to "can this process interact?" — stdin **and** stdout must be terminals. `check_tty()` uses it; `menu.py` imports it in place of its own `_can_host_a_menu`, which is removed. The two paths cannot drift again.
- `main.py`: one docstring line corrected ("stdin or stdout").

## How it is proven

- **Reproduced first**, on Linux with a pseudo-terminal as stdin and a pipe as stdout (the POSIX shape of Windows' `NUL`): `kitty egress` → exit 0, silent.
- **L1** (`tests/tui/test_prompts.py`): the asymmetric case for `check_tty` and for each prompt, which must refuse *without starting to draw*; a four-way table proving `check_tty` refuses exactly when both menus decline.
- **Real child process** (`tests/cli/test_stream_encoding.py`, next to the KBR-10 harness): `kitty egress` with an interactive-looking stdin and a piped stdout exits 2. The stdin is a pty on POSIX and `NUL` on Windows, so **all six CI legs** build the real asymmetry — before, `DEVNULL` meant Linux and macOS could not see this defect at all. Shipped with a premise test (the child really sees `True False`) and a falsification case that puts the old guard back and must observe the exact pre-fix signature (TEST_SUITE_IMPLEMENTATION_PLAN §1.4).
- **Test doubles** that meant "interactive" but faked stdin only now fake both streams (grep-verified, including one that stayed green either way).
- **Hand mutation sweep: 10 mutants, 10 killed** — stdin-only predicate, stdout-only predicate, `check_tty` bypassing the shared predicate, harness stdin falling back to `DEVNULL`, a prompt drawing before checking, menus back to stdin-only, runner dropping `stdin`, a refusal check that accepts anything, and the bait reverting both guards.
- 347 tests pass on the touched suites plus both layer guards; `ruff check`, `mypy src/kitty`, `lint-imports` clean. Line endings preserved per file.
- Reviews: design review (1 critical, found independently and fixed; 2 high, 3 medium, 3 low — all addressed or reasoned out) and code review (2 medium, 3 low → all applied → APPROVE).

## Docs

`TEST_SUITE.md` §8.2: child-interpreter count for that module 35 → 38 (measured), the stale "×4 Python versions" corrected to the six legs §8.4 added, runtime re-measured, and a note that a regression past both guards surfaces as a 60 s timeout rather than a fast failure.

## Noticed, not changed

`main.py::_run_setup`'s hint says "kitty egress does not need a terminal"; bare `kitty egress` needed one before this PR too (only `egress show` / `egress test` don't). Pre-existing and low; noted on KBR-204.

[KBR-204]: https://shelpuk.atlassian.net/browse/KBR-204?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[KBR-218]: https://shelpuk.atlassian.net/browse/KBR-218?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


shelpuk:
Addressed in 8f798c2 — the reviewer's docstring finding was correct.

`prompt_text`, `prompt_secret` and `prompt_confirm` each call `check_tty()`, which now refuses when **either** stream is not a terminal, but their `Raises:` lines still said "If stdin is not a TTY". All three now read "If stdin or stdout is not a TTY (see :func:`can_interact`)". I grepped `src/kitty` for any other stdin-only wording about the guard and found none; the remaining "non-TTY" mentions are generic and still accurate.

No behaviour change; `tests/tui/test_prompts.py` and `ruff check` are green. Line endings kept (CRLF).

PR #104, opened 2026-09-13
kitty bridge status and stop find the bridge start launched, on every platform (KBR-220)

shelpuk:
Closes [KBR-220](https://shelpuk.atlassian.net/browse/KBR-220). Also fixes a Windows defect found while investigating it: the background bridge died right after starting (ticket to be filed and linked; see "Jira" below).

**What this is about.** `kitty bridge start` runs the bridge in the background. `kitty bridge status`, `stop` and `restart` find it later through a small file, `bridge_state.json`, that the bridge writes when it is ready.

**What was wrong.**
1. **macOS, Windows, and Linux with a custom config folder:** the commands looked for that file in one folder, and the bridge wrote it to another. So:
   - every healthy `start` reported failure;
   - `status` always said "not running";
   - `stop` silently did nothing;
   - a second `start` launched a duplicate bridge.

   Plain Linux was unaffected only because the two folders happen to be the same there.
2. **Windows only:** the bridge crashed right after starting. It asked for a Ctrl+C/stop handler using a feature that exists only on Unix. So on Windows a background bridge could not stay up at all.

**What changes for users.**
- On every platform, `start` → `status` → `restart` → `stop` now manage the same, single bridge.
- A bridge started by a system service (systemd user unit, LaunchAgent) is also visible to `status` and stoppable by `stop`.
- On Windows the background bridge stays up, and the foreground `kitty bridge` no longer has the same crash.
- **Nothing to migrate.** The file stays where bridges have always written it (`~/.config/kitty/bridge_state.json`). A bridge already running when a user upgrades becomes visible to `status`/`stop` straight away.

## Decisions (product owner, 2026-09-13)

| # | Decision | Rejected alternative, and why |
|---|---|---|
| D1 | The state file lives in `~/.config/kitty` on every platform | The OS-specific config folder, next to `bridge.yaml`. More idiomatic, but bridges already running at upgrade would be orphaned, so it needs migration code. Accepted cost: on macOS/Windows the state file sits apart from `bridge.yaml`, as the bridge's keys file and logs already do. |
| D2 | One shared resolver **and** `start` tells the bridge the path explicitly (`--state-file`) | The resolver alone would still let the bridge ignore a path it was given, and that silent ignore is why no test caught this. The flag alone would not cover service-started bridges. |
| D3 | Ticket AC-4: nothing to migrate, recorded rather than coded | No kitty version ever wrote the state file to the other folder. The only file the old commands created there is a zero-byte lock. It is left alone, because deleting a lock an older kitty may hold is the riskier act. |
| D4 | Fix the Windows crash in this PR | Without it, ticket AC-2 (Windows CI leg) cannot pass: the bridge is dead before `status` runs. |
| D5 | Where the platform has no signal handlers, register none (no fallback) | On Windows `stop` ends the process outright, so no handler could run anyway. Ctrl+C still shuts a foreground bridge down cleanly. |
| D6 | The helper lives in `kitty.bridge`, not a new top-level module | Both callers already import `kitty.bridge`, so no new import contract is needed. |

## What changed

| File | Change |
|---|---|
| `src/kitty/bridge/state.py` | `default_state_path()`: the one resolver, read on each call |
| `src/kitty/bridge/manage.py` | Default comes from the resolver; `start_bridge` passes `--state-file` to the child |
| `src/kitty/bridge_runner.py` | Accepts `--state-file`, else uses the resolver; stop handlers via the helper |
| `src/kitty/cli/main.py` | `start/stop/restart/status` use the resolver (`bridge.yaml` still comes from the config folder); both foreground loops use the helper |
| `src/kitty/bridge/stop_signals.py` | New `install_stop_handlers`: registers SIGINT/SIGTERM where supported, returns `False` where not |
| `.system_design/SYSTEM_DESIGN.md` | **New, partial.** §1 Background bridge management: state-file contract, decisions, known limits |
| `.system_design/TEST_SUITE.md` | §8.2: the new real-process `l1` module and its cost (count 10 → 11) |
| `tests/test_layer_selection.py` | Registers the new `l2` source guard |

## Tests

| Test | Layer | Proves | Before the fix |
|---|---|---|---|
| `tests/cli/test_bridge_state_location.py`, journey: real `kitty bridge start` (exit 0, URL) → `status` (Running) → `restart` (new bridge recorded, old one gone) → `stop` (process gone) → `status` (not running) | l1, real processes | AC-1, AC-2, AC-3 | **Red** |
| same file, service route: real `kitty.bridge_runner --config` (no flag) → `status` sees it → `stop` ends it; POSIX exit code 0 (graceful) | l1, real processes | Service-started bridges; graceful SIGTERM | **Red** |
| same file: real runner with `--state-file` elsewhere writes there, not to the default | l1, real processes | The bridge honours the hand-off | — |
| `tests/bridge/test_bridge_state.py::TestDefaultStatePath` | l1 | Resolver ignores a different config folder | Red (no resolver) |
| `tests/bridge/test_bridge_management.py::…::test_start_bridge_tells_the_child_where_to_write_its_state` | l1 | `start_bridge` passes the path | **Red** |
| `tests/bridge/test_stop_signals.py` | l1 | Registers both signals where supported; declines, without raising, on a Windows-like loop | — |
| `tests/cli/test_bridge_stop_handlers.py` | l1 | Both foreground loops register handlers and shut down through `stop_async` | — |
| `tests/test_stop_handler_call_sites.py` | l2 | No code calls the Unix-only API directly | — |

**Isolation of the real-process tests.** They never touch a developer's profiles. Home and the platformdirs folders point at a temp dir: `XDG_*` on Linux/macOS, `WIN_PD_OVERRIDE_LOCAL_APPDATA` on Windows. Plain `LOCALAPPDATA` does *not* redirect Windows. A child first reports where kitty will look, and the fixture refuses before writing anything unless that is the temp dir. The config folder is deliberately not under home, so the regression test also fails on unfixed code on plain Linux. All `KITTY_*` variables are dropped, and the model catalog cache is pre-seeded so no network fetch happens during `start`'s 5 s window.

**Mutation sweep, by hand, in a copy of the tree (baseline green).** 13 mutants, all killed by their named test:
- **CLI reverts:** `status`, `start`, `stop` or `restart` alone, and all four together.
- **Hand-off:** the runner ignores `--state-file`; `start_bridge` drops the flag.
- **Resolver:** uses the config folder.
- **Helper:** catches the wrong exception.
- **Direct calls:** the runner, or a foreground loop, calls the Unix-only API itself.
- **Handler registration:** the runner registers none; either foreground loop registers none.
- **Isolation removed:** the fixture refuses in 0.8 s, before any write.

**Local (Linux):** `ruff check .`, `mypy src/kitty`, `lint-imports` (5 kept) are clean. Affected files plus both layer guards: 301 passed. The real-process module takes ~6.5 s. **Windows and macOS evidence comes from this PR's CI legs.**

## Review
- **Design review, round 1: "needs changes".** It found:
  - the untested "bridge honours `--state-file`" claim;
  - a zombie-process slowdown in the service test (11.6 s → 1.6 s);
  - a wrong claim of mine about platformdirs on Windows (verified in its source, corrected);
  - leaking `KITTY_*` variables;
  - an untested `restart`.

  All were addressed. The disposition is recorded in the requirements file.
- **Code review, round 1: "request changes".** It found:
  - nothing proved the foreground loops still register handlers (confirmed by mutation; test added, both mutants now killed);
  - a PID-reuse-sensitive restart check;
  - decision numbering drift;
  - two missing docstrings.

  All were addressed.

## Known limits, recorded in SYSTEM_DESIGN §1.5, not fixed here
- A Windows "background" bridge still shares the console it was started from, because `start_new_session` is POSIX-only.
- A Windows bridge is ended outright by `stop`, so no graceful shutdown or session summary.
- Services running under another account are outside the contract: the NSSM script runs as LocalSystem, and a LaunchAgent with `KeepAlive` restarts after `stop`.

## Other findings, not fixed here
- `tests/test_review_workflow_cli_contract.py::_kitty_env` sets only `LOCALAPPDATA`, which does not isolate Windows (see the isolation note above).
- `tests/bridge/test_bridge_state.py::_default_state_path` is pre-existing dead code.

## Jira
The Atlassian connector disconnected mid-session. The new Windows ticket, its link to KBR-220 and the implementation comment are pending reconnection.

[KBR-220]: https://shelpuk.atlassian.net/browse/KBR-220?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


shelpuk:
Thanks for the review. On the one suggestion, `tests/bridge/test_bridge_state.py:18-19` `_default_state_path`: agreed, it is dead. It is pre-existing, and this PR deliberately leaves it alone under the scope-discipline rule you cite. It is listed under *Other findings, not fixed here* in the description, so no code change here.

Separately, CI evidence for ticket AC-2: the Windows and macOS legs both collected the same 5,616 tests and passed. None of their skips is in the new modules. The `TEST_SUITE.md` §8.2 entry now records that result, in place of the per-leg timing it had promised: the gate does not print per-test durations.

PR #113, opened 2026-09-13
Recover when Anthropic rejects kitty's thinking history: strip only the broken thinking and retry (KBR-238)

shelpuk:

**What was broken:** a Claude Code session with thinking on, through kitty to Anthropic, could **stop answering outright**. Anthropic now verifies every piece of the model's earlier reasoning ("thinking") it is sent back. It rejects the whole request with a 400 if a thinking block has no signature, has a different one, or follows history that was edited. Kitty does all three:
- it drops signatures when it translates;
- it inserts unsigned placeholder thinking blocks;
- it trims long histories to fit the context window.

Nothing recognised the rejection, so the user got an error. In a load-balanced profile a healthy backend was also taken out of rotation. Filed as **KBR-238 (High)**.

**Proven live** against api.anthropic.com on `claude-sonnet-5`, `claude-opus-4-6`, `claude-fable-5-1` and, in manual thinking mode, `claude-haiku-4-5`. Through a real kitty bridge on current `main` (which now includes KBR-227), all four cases below returned the 400 to the agent; with this PR all four answered:
- `anthropic`, non-streaming;
- `anthropic`, **streaming**, which is how Claude Code always talks, so the second turn of any thinking session on that provider fails today;
- `custom_anthropic` native, non-streaming;
- `custom_anthropic` native, streaming.

**What this PR does:** when Anthropic rejects kitty's thinking history, kitty removes the thinking Anthropic can't verify and retries the same backend.

**Your decisions, recorded in the design doc:**
1. **Strip and retry** rather than change how kitty trims history.
2. **Strip only the broken blocks**, keeping the model's newer, valid reasoning. A live probe showed reasoning produced *after* a strip stays valid. Stripping everything would have erased it on every later turn, because Claude Code re-sends the full history and kitty re-trims it each turn.

**What it costs, stated plainly:**
- On a turn whose history is still broken, up to three rejected round-trips, and the model loses the earlier reasoning that was broken.
- A prompt-cache rewrite when the history switches between stripped and unstripped.
- The only operator signal today is a WARNING log line. A `/stats` counter is added to KBR-228's scope.

**Not in this PR:**
- Recovery for Codex, Chat Completions and Gemini *streams* (added to KBR-232).
- Stopping kitty from sending unsigned blocks at the source (KBR-228).
- Anthropic's beta `drop_block` setting, recorded as the alternative not taken: live it fixes edited history but still rejects missing or altered signatures, and it would add a beta header kitty doesn't send today.

## What changed

**`src/kitty/bridge/server.py`**
- **`_is_thinking_signature_error(status, body)`** recognises both live error texts on a 4xx.
- **`_strip_thinking_blocks(body, *, through_message=None)`** removes thinking blocks, copy-on-write:
  - with `through_message`: `thinking` and `redacted_thinking` in `messages[0..N]`, plus unsigned `thinking` anywhere;
  - without it: everything.

  A turn left empty keeps `content: []`, which is accepted live, so message indices never shift.
- **`_recover_rejected_thinking(body, error, strips_done)`** makes two strips targeted at the message the rejection names, then everything on the third, then stops. With no path, or a targeted strip that finds nothing, it strips everything at once.
- **`_make_upstream_request`** (non-streaming, all handlers) recovers before any path can blame the backend. Its count is per call, so per backend.
- **`_stream_messages`** recovers beside the issue-#32 repair:
  - Strips are counted **per serialized body**, so a failover's rebuilt body gets its own recovery.
  - The #32 repair is skipped for a stripped body, so the two can't take turns.
  - Strips give their attempt back to the retry budget, as transport-grace retries do.

**Register and design**
- **Register row M17** plus `Trigger.THINKING_SIGNATURE_REJECTED`, added to the corpus loader's refused triggers.
- **§3.2.1 M17** records the reasons, the live evidence, the costs and the alternative. §4.3 C3 now lists five body-changing paths. Counts are updated in the corpus README and the implementation plan.

## How it was verified

- **`tests/bridge/test_thinking_signature_strip.py`** (34 tests):
  - the detector against the verbatim live texts, as dict and as text, plus negatives;
  - the targeted and full strips, and the escalation and cap;
  - bridge flows, streaming and not: retry succeeds, later reasoning is kept, a pool keeps the backend healthy, a failover after strips recovers on the next backend, the cap surfaces the error, the #32 repair doesn't ping-pong, strips don't spend the stream's retry budget, and a Chat Completions upstream is left alone.
- **Planted-bug sweeps** in a copy of the tree with an absolute `PYTHONPATH`. The final sweep caught **20 of 20**. The two newest tests were each shown to fail when their fix is removed.
- **Reviews:**
  - `system-design-reviewer` raised 3 blockers. All were answered live and by your two decisions, and the per-body counting fixed a real quarantine bug it found.
  - `code-reviewer` raised a mypy failure, retry-budget exhaustion on streams, a missing corpus refusal test and stale counts. All fixed.
- **Checks:** `mypy src/kitty` clean; `ruff check` clean.
- **Wider run:** `tests/bridge`, `tests/harness` and the guards passed, 3,219 tests. After merging `main` (KBR-227/237), 608 related tests pass.

Jira: [KBR-238](https://shelpuk.atlassian.net/browse/KBR-238)

[KBR-238]: https://shelpuk.atlassian.net/browse/KBR-238?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


shelpuk:
Addressed in follow-up **#115**. The finding was valid: once the strips ran out, the `/v1/messages` stream's failover block, `_request_with_retry_balancing`'s failover loop and its final retry ladder all marked the backend unhealthy, breaking M17 and KBR-238 AC-1. #115 gates all three on `_is_thinking_signature_error`. The loop reads `last_exc`, because the tighter-compaction retry can replace the error. It adds pool-mode tests for streaming, non-streaming, the ladder and the compaction retry; each gate was hand-falsified, 5 of 5 runs. #113 merged before the fix was pushed.

PR #115, opened 2026-09-13
KBR-238 follow-up: a rejection that outlives recovery no longer quarantines a pool member

shelpuk:

**What was still broken after #113:** kitty now recovers when Anthropic rejects the model's saved reasoning ("thinking"), by stripping the broken blocks and retrying. But if that recovery ran out, a **load-balanced profile marked the Anthropic backend as broken** and took it out of rotation.

That is wrong: the rejected history is kitty's, so it would fail on any backend. In a pool, one broken conversation could cool down healthy backends and give other sessions "all backends unavailable" errors. It also broke KBR-238's first acceptance criterion ("without marking it unhealthy"). The automated review of #113 found it; #113 merged before this fix was pushed.

**What this PR does:** a rejection that recovery can't fix no longer marks any backend unhealthy.
- **Claude Code streams** (how it always talks): the error is shown to the agent at once.
- **Non-streamed requests:** kitty may try another backend, and stops after two rejections, as it does for any bad request.

**Cost:** none on the normal path. This only changes what happens after recovery has already failed, which is rare: the third strip removes all thinking.

## What changed

**`src/kitty/bridge/server.py`** has three places that mark a backend unhealthy. Each now skips a thinking-signature rejection:
- **the `/v1/messages` stream's failover block:** the error then reaches the agent, since a 400 is not retried;
- **`_request_with_retry_balancing`'s failover loop:** it reads `last_exc`, because the tighter-compaction retry may have replaced the original context-too-large error, and compaction is what breaks signatures;
- **its final retry ladder.**

**`.system_design/TEST_SUITE.md` M17:** now says what happens once recovery runs out. A non-streamed pool picks a backend again, possibly the same one, since none was cooled.

## How it was verified

**4 new tests** in `tests/bridge/test_thinking_signature_strip.py`:
- **a pool whose recovery runs out:** streaming and non-streaming, every backend stays healthy;
- **the final retry ladder:** written to hold whichever backend the random pick chooses; an earlier version failed about half the time, and code review caught it;
- **the tighter-compaction retry:** a signature rejection there cools nothing.

**Planted bugs:** each of the four changes was removed in turn, and every one was caught in 5 of 5 runs.

**Checks:**
- `tests/bridge` + `tests/harness`: 3,040 passed.
- `mypy src/kitty`, `ruff check`: clean.

**Review:** `code-reviewer` requested changes: a flaky test, overstated failover wording in the docs, and the `exc`/`last_exc` mix-up. All three are fixed.

**Noted, not changed:** the F4 "two 400s" counter just below the new gate still reads the original error rather than the retry's. That line predates KBR-238 and only affects when failover stops, so it is low priority.

Jira: [KBR-238](https://shelpuk.atlassian.net/browse/KBR-238)

[KBR-238]: https://shelpuk.atlassian.net/browse/KBR-238?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


shelpuk:
**Session follow-up (2026-09-14): merge currency, two conflict rounds, and the KBR-232 seam fix

**1. Merged origin/main twice** (16 + 9 commits landed mid-flight: KBR-224/225/226/230/241, then KBR-228/240/232 + docs). Both spec conflicts resolved by carrying both sides: KBR-228's M16/M17 amendments (system-carrier restore, `/stats thinking_stripped` counter) and KBR-232's handler/recovery wording, with this PR's expanded no-quarantine passage and the four→five body-changing-path fixes on top. The diff vs main remains exactly the intended delta.

**2. Re-established all evidence on the merged trees**: 38/38 in `test_thinking_signature_strip.py`; the four follow-up tests re-falsified against un-fixed main (each fails at its named assertion: backend health map / `assert 429 == 400` / recorded failure counts); register, answered-questions and layer-marker guards; mypy and ruff clean; the full bridge suite locally — 1,753 passed on the exact pushed bytes.

**3. Fixed the KBR-232 seam (owner decision: fix in this PR) — commit 3a93f32.** KBR-232's three translated stream handlers (`_stream_responses`, `_stream_gemini`, `_stream_chat_completions`) had unguarded balancing marks: a signature rejection that outlives the strip budget cooled a pool member there — the same defect this PR fixes for the Messages stream, and a contradiction of the merged M17 row. All three marks now carry the same `_is_thinking_signature_error` exemption; three new pool tests (one per protocol) fail at the health-map assertion against the pre-guard server and pass with it. The CI review's orphaned Critical (server.py:3289) was exactly this gap; the 13:05 review verified it resolved — `No findings`, `has_blocking: false`.

**4. Accepted-latent seam, flagged for the owner** (unchanged, recorded on KBR-238): a signature rejection arriving as a pre-emission 200+SSE error event follows KBR-241's ladder rather than M17's immediate surfacing; live delivery is always status-4xx today.

State: all checks green on 3a93f32 (`ci-required` with every leg completed; review SUCCESS), MERGEABLE/CLEAN. Ready for the owner to merge — not merged from here.

PR #117, opened 2026-09-13
Forward the agent's own thinking budget_tokens instead of deriving it from max_tokens (KBR-225)

shelpuk:

When an agent such as Claude Code talks to an Anthropic-compatible upstream **through kitty's translated route**, kitty has been throwing away the "thinking budget" the agent chose (how many tokens it may spend reasoning) and substituting one computed from the request's `max_tokens`. Anthropic bakes that budget into the prompt it caches, so any turn whose `max_tokens` changed gets a cache miss and pays full input price again — kitty's own rewrite was creating that bill, even though the agent had asked for the same budget on both turns.

After this PR, the agent's budget is forwarded untouched whenever it is valid, so the prompt cache survives a `max_tokens` change. If the agent sent no budget — or an invalid one — kitty keeps today's behaviour unchanged as a documented fallback. This is one brick of the larger KBR-197 epic (prompt-cache fidelity): today the cost was mostly hidden because another change strips the cache markers; once that epic restores them, this fix is what stops the restored cache from silently missing.

## What changed

- `MessagesTranslator.translate_request` carries the agent's `thinking.budget_tokens` on the internal key `_thinking_budget_tokens` when it is valid (an int, `>= 1024`, `< max_tokens`; a non-int `max_tokens` is treated as absent so malformed input still errors exactly where it did before).
- `AnthropicAdapter.translate_to_upstream` ships the carried budget verbatim and leaves `max_tokens` alone on that branch (a valid budget already implies `max_tokens >= 1025`, so the old raise cannot trigger); the fallback branch keeps the pre-fix derivation byte-identically.
- `_thinking_budget_tokens` joins `_INTERNAL_KEYS` and the completeness guard's expected set, so it can never leak to a non-Anthropic wire.
- KBR-203's R1/R2 tests pinning the defect are inverted per their docstrings; new translator negative matrix, adapter-level negative controls, and a characterisation pin at the `max_tokens = budget + 1` floor.
- TEST_SUITE.md §3.2.2 row P5c rewritten (forward branch recorded as trigger-met-but-inert, M15 precedent) and the §3.3.2 anchor passage updated; `tests/harness/register.py` P5c comment moved with it.

## Verification

- The four pre-fix reds were watched failing for the right reason before the fix; all green after.
- Full l1+l2 gate (CI's command): **5997 passed, 2 skipped**.
- Layer/register guards (266) and internal-key guards (111) green; ruff + mypy clean.
- Mutation spot-check in a throwaway copy: four hand-applied mutants (validity floor dropped, upper bound dropped, key unregistered, forward branch disabled) each killed by exactly the tests built for them.
- Reviewed by the system-design-reviewer (spec) and code-reviewer (APPROVE) before push.

Closes KBR-225. Part of epic KBR-197.


shelpuk:
## Merge conflicts resolved (two rounds — main moved twice while this PR waited)

Main landed **four** PRs against this branch's base: #116 (KBR-226), #118 (KBR-224), #119 (KBR-230) and #121 (KBR-241). Merged all four in; history is `0d44054` (KBR-224+226+230) then `078bc8c` (KBR-241).

**Round 1 — KBR-224/226/230 (`0d44054`).** KBR-224 (`output_config`) works the *same translated route* as this PR, touching six of its seven files, so this was a semantic merge, not just a textual one:

- Code auto-merged and verified correct: the `_output_config` carry sits beside the `_thinking_budget_tokens` carry in `MessagesTranslator.translate_request`; both keys are in `_INTERNAL_KEYS`; the adapter restores `output_config` after the thinking branches; the `forwards_output_config` per-adapter flags landed intact. The two features are independent by construction — one carries `thinking.budget_tokens`, the other top-level `output_config` — and both PRs' merged test suites pass together (676 tests, including KBR-224's new `test_anthropic_output_config.py` and KBR-226's parallel-tool tests).
- `TEST_SUITE.md` §3.2.2 was the only textual conflict, and it needed judgment: both PRs edited row P5c. Resolution keeps **this PR's rewritten P5c row** (forward branch, trigger-met-but-inert, M15 precedent) with **KBR-224's M9-fallback sentence folded in** (the native-retry converter now also lacks `output_config`) and the closing translator-only-keys sentence extended to name `_output_config`. KBR-224's P5d row taken unchanged.

**Round 2 — KBR-241 (`078bc8c`).** Textually clean: KBR-241 is response/streaming-side (preamble hold, pre-emission error recovery) and edits a different region of the design doc; its `server.py` diff never rebuilds a request body, so the budget-forwarding branch is unaffected. Verified, not assumed.

Verification after both merges: 679 tests across the touched provider/bridge/streaming files, layer+register guards (266), internal-key guards, ruff and mypy — all green.

PR #121, opened 2026-09-14
KBR-241: recover pre-emission upstream error events on Messages-wire streams

shelpuk:

When the upstream provider failed early in a reply with its own error — the `API Error: Server error mid-response` in KBR-241's log — kitty passed it straight through to Claude Code, which stopped your turn and waited for you to type "proceed", even though simply re-asking the provider would have succeeded. This landed kitty in its most expensive failure mode: the agent halted, and the recovery cost a human.

Now kitty does the re-asking. An upstream error event that arrives **before any content** is judged before the client sees it: the attempt is retried on the same ladder as an empty reply (same-backend backoff, failover on a balancing profile), and only when every attempt fails does the agent hear about it — receiving the **provider's own error** (so `overloaded_error` stays `overloaded_error`), not kitty's wording. You typed "proceed" one fewer time per incident; on a balancing pool, a sick backend is routed around automatically.

**Scope, per your 2026-09-14 decision (recorded in TEST_SUITE.md §11):** pre-emission only. After content has reached the client, behaviour is unchanged (Q14(a) stands) — there, the agent's own retry remains the only recovery that cannot replay text or splice tool calls. KBR-233 (same defect, closed "NOT fixed") required exactly this kind of fresh owner decision before revisiting; KBR-241 supplied it, and this PR is linked to KBR-233 so the lineage is on record. KBR-233's acceptance criteria (CC-wire-style quarantine) were **not** executed: the health model of the empty-reply ladder is kept deliberately — retry without quarantine — and the §11 amendment owns that difference rather than erasing it.

## What changed

- `PreambleHold` (`src/kitty/bridge/preamble_hold.py`): an `error` event before content no longer releases the hold. It is recorded (`error_seen`, `error_event`, `error_event_complete`) — recognition unchanged (SSE name or `data.type`, decided at the name line, as the official SDK decides). Completion keys on the event's terminators (blank line / next header), so a chunk boundary between the name line and its data line cannot truncate the payload. The judge is deaf to anything post-error: a later `message_delta` stop reason records nothing, making the answer independent of chunk framing.
- `_stream_messages` (`src/kitty/bridge/server.py`, native Messages-wire branch — native adapters **and** translated Anthropic-wire routes via `_serves_messages_wire`): stops reading at the completed error event and runs the ordinary empty-response ladder. At exhaustion: `502` with the provider's payload re-embedded plus `reason: "upstream_error"`; when the payload carries no usable `error` object, D4's `reason: "empty_response"` body. Discarded attempts log `upstream_error=<type>`, are not counted as completions, and do not reset backend health. The guarded-dead post-emission arm ends an already-open stream with the provider's payload raw.
- Records: TEST_SUITE.md §11 Q14(b) D2 amendment (owner decision, KBR-233 supersession, health-model and exhaustion-shape ownership, three accepted residues) plus the two dependent passages; README troubleshooting gains the `upstream_error` 502; module/method docstrings updated.

## Testing

- 19 tests added or rewired across `tests/bridge/test_preamble_hold.py` (the judge, L1) and `tests/bridge/test_native_empty_stream.py` (the whole handler against real sockets and scripted upstreams, L1/L2): **90 pass** in the two files, including flipped D2 pins, exhaustion bodies (both payload shapes), D3-vs-error ordering in both framings, the post-emission passthrough with completion accounting, log line + negative control (exhausted), and real-socket stall tests that fail fast on regression.
- Falsification: a 7-mutant sweep in a pristine copy (fresh tree + pyproject, green 90-test baseline) — **all mutants killed**, including the completion-signal bug the review round caught (killed by the new bare-data completion tests) and "handler never breaks on error" (killed by the all-attempts-stall ladder, 6×2 s vs the instant break).
- Gates: full PR gate **6006 passed, 2 skipped** (20m48s); `ruff check .`, `mypy src/kitty`, `lint-imports` clean.
- Review rounds: system-design-reviewer ×2 (requirements rebuilt on its findings; both rounds' findings resolved), code-reviewer ×2 ending in **APPROVE** with independent re-falsification of the fixes.


shelpuk:
Addressing the automated review's four suggestions:

1. **Usable-payload duplication — fixed** (2f0d4f1). The predicate now lives in one place, `_usable_upstream_error_payload()`, shared by the discarded-attempt log, the guarded `sr` arm, and the exhaustion body.
2. **Class Attributes omission — fixed.** The `PreambleHold` class docstring now names the `error_seen`/`error_event_complete` properties and why the caller waits for the latter.
3. **Conditional format-arg — fixed.** The WARNING's suffix is hoisted into `error_note` before the call.
4. **The `sr is not None` test — declined, deliberately, on the record.** The arm is guarded-dead post-KBR-183: a released attempt breaks the attempt loop, and every post-emission exception path with a prepared response finalizes and breaks, so no end-to-end test can reach it without mocking the internals the repo's suite deliberately avoids. This was scoped in REQUIREMENTS AC-5 and the PR description as reviewed-but-untested; the threading itself (one terminal SSE `error` event, provider payload raw, no `message_stop` after) was verified by the code review. If the team prefers a seam to make it testable, that's a follow-up decision — happy to file it.

Tests re-run after the fixes: 90 pass in the two touched suites, ruff/mypy clean; the platform checks re-run on the new head.

shelpuk:
Addressing the second automated review's two warnings (head `5272078`):

**1. Errored-but-unusable exhaustions no longer report `empty_response` — fixed.** Agreed, and this was a flaw in my round-1 spec choice, not just the diagnostics: the ladder *watched the upstream error*, so reporting an empty reply is false reporting against D4's own rationale for the marker. Now: a usable payload is delivered as before (provider's error + `reason: "upstream_error"`); an unusable one gets kitty's own wording — *"Kitty Bridge received an error from the upstream provider before any content, on every attempt, but the provider's error payload could not be delivered."* — with the same `reason: "upstream_error"`. Only a genuinely empty ladder reports `empty_response`. The discarded-attempt log marker likewise always reflects a recorded error: provider `type` when extractable, `unknown` for a typeless-but-usable payload, `unusable` for a malformed one. TEST_SUITE.md §11's amendment, README, and the requirements record all move with the behavior; the flip is pinned by `test_exhaustion_with_an_unusable_payload_still_reports_the_upstream_error` and the marker parametrization (`well_formed` / `name_only_unparseable` / `type_only_malformed` / `unknown`).

**2. The stall test's vacuous health loop — fixed.** Correct on both counts: the single-backend server's `_backend_health` is empty (the loop never ran), and the len/status/body assertions pass on a timeout-path regression. The test is renamed to what it actually pins (`..._still_serves_the_retry`), the dead loop is gone, and its docstring states plainly that the kill for a read-past-the-error build lives in `test_every_errored_attempt_ends_on_its_error`.

Falsification re-run after the changes: 93 tests pass in the two suites; the 8-mutant sweep (now including `handler_unusable_keeps_empty_response`, encoding warning 1's original behavior) kills all mutants against a green baseline. A fresh CI round runs on the new head.

shelpuk:
Third CI round red — root cause found and fixed in `f2a191e`. `TestNoUnreviewedVendorToken` correctly refused the new `_NATIVE_UPSTREAM_ERROR_MESSAGE` literal: every product-named string in a request-building module must be classified in `_ALLOWLIST` with proof it cannot reach an upstream request. Classification added beside its KBR-155 siblings: the constant is built in `_stream_messages`' native branch *after* the upstream reply has ended and is written only to a downstream 502 body or an SSE error event on `sr` — no `build_upstream_headers()` or body path can reach it. Guard test + both touched suites green locally (102 passed); fresh CI round runs on `f2a191e`.

PR #122, opened 2026-09-14
KBR-240: give each Responses output item its own output_index

shelpuk:

When you use Codex CLI with kitty on a Chat Completions model (MiniMax, DeepSeek, etc.), kitty translates Codex's "Responses" protocol to the provider's wire. A normal Codex exchange is: the model writes a short sentence, then calls a tool. Before this fix, the translated stream announced **both** the sentence and the tool call at the same positional slot (`output_index: 0`) and closed that slot twice — every event was valid on its own, but the sequence was grammatically wrong, and any client that positions content by slot could merge or misplace the two. Codex happens to cope by tracking item ids; the wire itself was malformed, the same class as KBR-226 (fixed for Claude Code's wire two days ago).

This PR gives each output item its own slot. It also fixes two adjacent holes the ticket's own audit clause flushed out on the "upstream died mid-stream" path (timeouts, dropped connections): the model's thinking item was never closed and the final summary could silently drop items. And it adds a field OpenAI's own grammar requires on tool-argument events but kitty never sent.

**Found during KBR-226's sibling check, filed as KBR-240, fixed here.** A second, separate defect was discovered by this PR's own server-level test — the translated `/v1/responses` stream never opens with `response.created` (the server builds those events and never writes them). It is **not** fixed here; filed as **KBR-242** and recorded as gap G41 + `SYSTEM_DESIGN.md` §4.5.

## What changed

- `src/kitty/bridge/responses/translator.py` — one shared output-item counter, allocated when an item opens (reasoning, text, each function call); every event referencing an item carries its recorded index, on the chunk path, the finish path, and the EOF-fallback path. The upstream CC `tool_calls[].index` is now a routing key only. `synthesize_completed_events` closes the reasoning item and started-but-stripped-empty text items; both closing paths include every opened item in `response.completed`'s `output`, sorted by index, so array position equals `output_index` throughout.
- `src/kitty/bridge/responses/events.py` — `output_index` (vendor-required) added to `response.function_call_arguments.delta` / `...done`.
- Design docs: `SYSTEM_DESIGN.md` §4 (translator stream-item positioning — new section), `TEST_SUITE.md` G40 (closed) / G41 (open, KBR-242), implementation-plan §16 row.

## Evidence

- **Red at base, watched**: all 10 L1 tests + the server-level walk fail against the base revision (11 failed), pass after the fix; watched again as a whole after the review amendments.
- **Mutation checks**: a constant-0 counter mutant is killed by all 9 stream tests; a mutant pinning every text-item event site to 0 is killed by 3 tests (the gap the design review caught in the first draft).
- **Full fast gate** (`pytest -m "l1 or l2"`): **6,033 passed, 2 skipped** on the final tree.
- **Reviews**: system-design review and code review both passed (the code review independently replicated the mutation kills); both required an A2-coverage fix and a synthesize-reasoning fix, which are in this PR.
- Layer guards: `test_layer_markers.py` + `test_layer_selection.py` green; new test file defaults to l1.


shelpuk:
Both suggestions from the automated review are applied in 96a6e77:

1. **Synthesize-path regression added** (`test_synthesize_closes_text_stripped_to_nothing_before_a_call`): thinking-tag-only text, then a tool call, then EOF. Red at the `origin/main` revision, and it is the mutation evidence itself — a mutant reverting the synthesize close gate to the old `clean_text` predicate is killed by this test and nothing else in the suite (verified: 1 failed, 10 passed under the mutant). My first attempt at that mutant hit the wrong one of the four `_text_started` gates (the finish path's array gate — separately pinned, which that accidental run confirmed), so the kill was re-verified against the intended third occurrence.

2. **Doc citations fixed in both files** — the new walk now cites `.system_design/SYSTEM_DESIGN.md` §4.2 / G40, and the KBR-226 sibling's identical drift (`test_parallel_tool_use_stream.py` line 12) now cites `TEST_SUITE.md` §6.2.2 / G39, since this PR was flagged as the natural place to fix the pattern.

Local gates re-run before push: ruff, mypy, import-linter, and all four touched test files (85 passed).

PR #127, opened 2026-09-14
KBR-243: hold agent requests while backend recovery fits inside 300s

shelpuk:

When every LLM backend behind a kitty profile is temporarily unavailable — most commonly rate-limited — kitty used to answer the agent's request with an immediate 503 ("All 4 backends are currently unavailable … ready in 114s"). Claude Code treats that as a fatal error and **abandons the turn**, so the developer's session stalls until a human notices and re-sends the prompt. The ticket's incident lost a session to an outage that was ~2 minutes from self-healing.

Kitty now **holds such a request at the front door** while at least one backend will recover within 5 minutes (the window the ticket specifies): it waits out the cooldown quietly, then flies the request normally. The developer just sees the answer arrive a minute or two later. If nothing will recover within 5 minutes, the same immediate 503 as today is returned, so worst case is never worse than before. This covers all four protocols kitty serves (Anthropic Messages, OpenAI Chat Completions, OpenAI Responses, Gemini).

## What changed

- `BridgeServer._select_backend_or_hold()` — new arrival-path selection used by all four protocol handlers: holds while the soonest cooldown expiry falls strictly inside 300 s of arrival, re-selects after each wait, and only then gives the today-shaped 503 (built from the *latest* cooldown information).
- Safety properties, each with a test: the wait never blocks other sessions (async sleep), a disconnected client ends the hold without firing a paid upstream request (transport poll, the KBR-241 preamble-hold pattern), the request body is drained first so the client is not stuck mid-upload, concurrent holds are jittered 0–2 s so a herd does not stampede the one recovering backend, and the fabricated "no stream-capable backend" 300 is never waited on (`recoverable=False`).
- `AllBackendsUnhealthyError` gains a keyword-only `recoverable` flag; `/stats` gains a `recovery_holds` counter; WARNING at each hold start, INFO when a hold ends in success.
- Mid-ladder failover paths are deliberately untouched (post-emission rule, Q14/KBR-163).
- Docs: `.system_design/SYSTEM_DESIGN.md` §6 (new — the health/cooldown machine and every decision's why) and the corresponding `TEST_SUITE.md` L3 row.

## Tests

- `TestRecoveryHold` — 12 unit cases on a stepped fake clock: hold-then-succeed, beyond-window, the strict 300 boundary, non-recoverable, repeated holds to exhaustion (latest exception returned), auth-sized cooldown, both gone-client shapes, jitter+clamp exactness, stats count.
- `TestArrivalHoldWiring` — HTTP against the real app: hold-then-200 for all four protocols, plus per-route 503s that prove the post-hold 503 names the latest cooldown.
- Existing 503-envelope fixtures moved beyond the window (264/120/120/180/90/196 → 400); the two 300 fixtures stay as HTTP boundary pins.
- Mutation spot-check (copied tree, absolute `PYTHONPATH`): 5 mutants — strict-window flip, dropped `recoverable` guard, sabotaged sleep, removed transport poll, dropped clamp term — all killed; green baseline verified first.

## Review process

Two rounds of system-design review (15 findings, all applied — including the client-disconnect semantics and the uniform-treatment-of-cooldown-kinds decision) and a code-review pass (2 major + 5 minor/nit findings; both majors and all accepted minors fixed; one optional hardening declined with reason in the commit message). Local runs: target file, balancing/circuit-breaker/stats suites, layer guards — green; the full `l1 or l2` gate is CI's to confirm.


shelpuk:
Addressed the review in 9c05c22 — findings, verified then fixed:

1. **Docstring vs drain placement (server.py)** — agreed: the drain sits inside the loop, so it runs before *each* hold, not just the first. Docstring, `SYSTEM_DESIGN.md` §6.2 and the task requirements now say "before each hold (aiohttp caches the read, so later iterations are no-ops)". Behaviour unchanged — aiohttp's `_read_bytes` cache makes the second and later `read()` calls no-ops.

2. **Mid-loop `recoverable=False` transition (tests)** — added `test_mid_loop_recoverable_transition_stops_holding`: first raise recoverable (one 100 s hold), second raise non-recoverable → the loop must return it at the next check with exactly one sleep.

3. **Near-expiry gamble integration (tests)** — added `test_truncated_wake_exits_through_the_near_expiry_gamble`, exercising the *real* truncation path end-to-end: `failed_at` 0.6 s before arrival makes `remaining` truncate 99.4 → 99, so the wake lands 0.4 s before the cooldown truly expires, and the re-selection must exit through the 60 s gamble branch (asserted: one 99 s sleep, then success — no second raise). This pins the §6.2 sentence the finding cites.

4. **README (orphaned finding)** — agreed it's a real gap: the hold is user-visible. Added a "Held while recovery is near" paragraph to the Balanced Profiles section documenting the 5-minute hold, the immediate-503 fallback beyond it, and the disconnect behaviour.

All 55 tests green locally; ruff clean.

shelpuk:
Fixed in 6c733a7: the class docstring now carries an `Attributes:` block naming `backends`, `retry_after` and `recoverable`, with the recoverable contract stated at the class level — today only the no-stream-capable raise sets False (its 300 is fabricated, not a recovery time), and the arrival hold sleeps only on recoverable raises. 55 tests green, ruff clean.

shelpuk:
Fixed in 2f82aa4: the README's `/stats` section now has a `recovery_holds` paragraph in the same per-counter style — what a hold counts (each held wait, a request held twice counts twice), and how to read it (rising = providers throttling; high with no failed requests = the hold doing its job).

PR #138, opened 2026-09-14
KBR-192: enforce XOR pairing of Image.digest and Image.ref

shelpuk:
## Context for Product Owner

`Image` declared a two-field invariant — `digest` is set when `ref` is not — that the rest
of `contract.py` already enforces for `Opaque.kind`, `Reply.stop_reason`, `Turn.role`, and
`Conversation.sampling`. `Image.__post_init__` only froze `cache_control`, so today it
constructs with both fields set, or with neither, silently. The ticket identifies the
no-identity case (an `Image` projects every image identically — the same blindness
KBR-179 names for `Opaque`) and the both-set case (two readers could populate the pair
differently for one image and show a phantom delta on content neither altered).

This change closes the gap on the test-suite side. Three reader sites already produced a
legitimate "neither" case under TEST_SUITE.md §7.4 rule 7 — Gemini's undecodable base64,
Gemini's missing/wrongly-typed `fileData.fileUri`, and Anthropic's missing/wrongly-typed
`url`/`file_id`. Per the 2026-09-14 owner decision, all three now keep the part in the
projection with *some* identity: the raw encoded bytes (a second `image_digest` recipe,
pinned by `TestImageDigestRecipe`) for the undecodable case, and `opaque_digest`'s
canonical-JSON recipe (already pinned for unmodelled blocks) for the others. The residual
still records the bad/missing value, so the run fails visibly at the right path.

## What changed

- `tests/harness/contract.py` — `Image.__post_init__` raises `ValueError` on neither/both;
  `image_digest` and the `Image` class docstring document both recipes and the XOR invariant.
- `tests/harness/reader_gemini.py` — `_read_inline_data` digests raw encoded bytes on decode
  failure; `_read_file_data` digests `opaque_digest(value)` when `ref` is missing, with
  the missing-key residual recorded explicitly.
- `tests/harness/reader_anthropic_messages.py` — `_read_image` url/file branches use the
  same canonical-JSON recipe when `ref` is missing.
- `tests/harness/test_contract.py` — new `TestImagePairing` (4 tests) and
  `TestImageDigestRecipe` (1 test pinning whitespace sensitivity).
- `tests/harness/test_reader_gemini.py` — three existing test expectations updated; new
  missing-fileUri test pins the run-fails-visibly behaviour.
- `.system_design/TEST_SUITE.md` — §3.3.1 line 599 mentions the raw-bytes recipe; §7.4
  rule 7 row 2 splits to keep `tool name` unchanged while adding `fileData.fileUri`; row 3
  rewords to describe the raw-bytes outcome; §11 records the 2026-09-14 owner decision.

## What does NOT change

- `Opaque`, `Reply`, `Turn`, `Conversation` post-init validations are untouched. Existing
  tests for those types continue to pass.
- The behavioural product test (`tests/providers/test_*`) is its own ticket (the KBR-197
  epic); the harness's L1 contract is what this change scopes.
- The reconciliation ticket for `reader_anthropic_messages.py:599` raising on undecodable
  base64 vs `reader_gemini.py` residualising (§7.4 "Reconciliation owed") is unblocked
  by this change but out of scope — both branches now adopt a compatible identity recipe.

## Verification

- 6275 / 6275 tests pass (5 skipped, 34 deselected); 1392/1392 harness tests pass.
- `mypy tests/harness/reader_gemini.py` clean.
- `ruff check` clean on all changed files.
- `lint-imports` 5/5 contracts kept.
- Falsification: XOR check disabled in scratch copy → both pairing tests fail; restored.

Closes KBR-192.


shelpuk:
Thanks for the review — all three suggestions evaluated; none applied. Reasoning per finding:

**1. Raw-bytes recipe pin only covers one wrapping shape (`test_contract.py:1689`)** — declined.
The test pins the *claim* that wrapping changes the digest: one wrapped shape against one unwrapped shape proves that claim. A second wrap position would add no new property — any wrap differs from unwrapped by the same mechanism. The pin exists so a future translator stripping whitespace fails loudly; one counterexample is what makes it loud.

**2. fileData digest hashes raw `value` rather than aliased view (`reader_gemini.py:1354`)** — declined; the current behaviour is the correct half of this choice.
Hashing the raw wire value means two blobs differing only in unknown junk keys digest **differently**. Hashing the aliased view (which drops unmapped keys into the residual) would make those two blobs digest **identically** — the exact blindness this ticket closes for `Opaque` (KBR-179) and `Image` (KBR-192). `opaque_digest`'s established contract operates on the block as the wire carried it; the fileData use keeps that contract. A wire-alias spelling difference surfaces in the residual path, which is where the run fails visibly anyway.

**3. Image class docstring could name the four cases the XOR admits (`contract.py:306`)** — declined.
The docstring states the rule once, both directions ("digest is None iff ref is not None"), and points at the enforcement — the same shape as the `Reply` docstring this one deliberately mirrors. Enumerating two-legal/two-illegal as four bullet cases would restate what `__post_init__` enforces and the two control tests pin.

CI: 12/13 checks green; the Windows 3.12 leg is still running.

PR #143, opened 2026-09-14
KBR-251: image-reader reconciliation — conform both shipped readers to T-A4's recipe

shelpuk:

Kitty Bridge's test suite proves message fidelity with a set of independent "readers" — one per provider wire format (Anthropic Messages, OpenAI Responses, Gemini) — that translate every request into a common form so the suite can compare what the agent sent with what the bridge forwarded. When an image arrives with malformed base64 (real traffic: standard base64 output is line-wrapped and the strict decoder rejects every line break), the three readers disagreed on what to do with it. One crashed the whole check, one silently dropped the image and shifted every later item's position — both failure modes would hide real bridge defects behind harness noise. This change makes all three readers handle the malformed image the same way: record the problem at the exact field that caused it, keep the image in its place with a deterministic identity, and let the check continue. One shared defect class closed; the suite can now trust a clean run on image-bearing sessions.

## What changed

- `tests/harness/reader_anthropic_messages.py::_read_image` — the base64 branch no longer raises `UnreadableBodyError` on undecodable payloads. It residualises the `…source.data` leaf and projects the part under `image_digest(raw.encode("utf-8"))`, the recipe KBR-192 pinned; media type and cache breakpoint carry as before. Missing/wrongly-typed `data` keeps today's behaviour (escapes as `KeyError`/`TypeError`) — explicitly out of scope.
- `tests/harness/reader_responses.py::_read_image` — the three `return None` drop paths are gone; the function now always returns an `Image`:
  - decode-fails data URL → residual at the `…image_url` leaf, digest of the payload string, declared media type carried;
  - non-base64 `^data:` URL → residual at the leaf, digest of the payload after the first comma, media segment parsed from the URL prefix (keeps `image_digest`'s "media type excluded" rule);
  - no `image_url` and no usable `file_id` → `opaque_digest(entry)` identity, mirroring `reader_gemini._read_file_data`'s missing-`fileUri` shape; the residual names the absent key as `None` and any wrongly-typed `file_id`.
  - a wrongly-typed `image_url` residualises even when a usable `file_id` carries the part — the malformed field cannot hide inside a ref-only projection;
  - a local `_residualise` helper (mirrored from the Anthropic reader) sweeps unmapped `input_image` siblings (`detail`, future keys) on every branch.
- `.system_design/TEST_SUITE.md` §7.4 — the "Reconciliation owed" note is paid: date, the two readers, the now-shared recipe, and an explicit list of what stays owed (Anthropic's missing/wrongly-typed `data`; the Responses `input_file` sibling sweep; the `_DATA_URL` regex's empty-media-segment and case-variant defects).
- `reader_gemini.py` is untouched — it was already the reference implementation.

## Tests

- New `TestUndecodableImagePayloads` in both reader test modules (10 tests): the recipe shape, part-keeps-position (index claim asserted for all three Responses failure shapes), sibling sweep on failure and success paths, mixed wrongly-typed-`image_url`+`file_id` case, and `verify_total` failing at the exact residual path.
- Updated `test_a_data_url_that_is_not_base64_residualises_rather_than_becoming_a_ref` to the new leaf-residual shape; `test_an_image_given_only_by_file_id_carries_the_id_as_its_reference` drops its `detail` key (a pre-existing silent drop, out of scope here).
- All 10 new tests were shown to fail against the old behaviour (red-first), and a mutation pass reverted each new branch in turn — every test killed its own mutant.

## Gates

- `ruff` clean · `mypy src/kitty` clean · `lint-imports` 5/5 kept · `pytest -m "l1 or l2"`: **6393 passed, 5 skipped, 0 failed**.

## Ticket trail

- Implements [KBR-251](https://shelpuk.atlassian.net/browse/KBR-251); unblocked by KBR-192 (Done, PR #138).
- [KBR-196](https://shelpuk.atlassian.net/browse/KBR-196) — earlier filing of the same reconciliation, prose predating KBR-192 — consolidated here via a Relates link and a comment; to be closed as "consolidated, NOT fixed" when this merges.
- Pre-existing defects observed but not fixed here (candidates for separate tickets): `input_file`'s silent sibling drop (`detail`), and the `_DATA_URL` regex rejecting `data:;base64,…` and `;BASE64,` case variants — both named in the §7.4 note.

[KBR-251]: https://shelpuk.atlassian.net/browse/KBR-251?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


shelpuk:
Fixed: the new-test tally is **10** (2 Anthropic + 8 Responses), not 9 — counted before the clean-success-path sibling test was added. PR description and the Jira ticket comment are now aligned. Thanks for the catch.

PR #145, opened 2026-09-14
KBR-218: ask the console API on Windows, fail closed

shelpuk:
# KBR-218 — On Windows, an interactive command with stdout redirected to NUL still passes the terminal guard

Closes [KBR-218](https://shelpuk.atlassian.net/browse/KBR-218). Parent epic: KBR-123 (Terminal and platform compatibility).

## Context for the product owner

**What a Windows user saw.** `kitty auth openai > NUL`, or any parent process (a CI
runner, a wrapper script) that starts an interactive kitty command with both its
input and its output pointed at the Windows null device, crashed with a Python
traceback (`NoConsoleScreenBufferError`) instead of the intended message
"kitty … requires an interactive terminal (TTY)" and exit code 2. Same user-facing
family as KBR-10, KBR-187 and KBR-204 — kitty misbehaving on Windows because its
streams are redirected — reached by one more route than any of those tickets closed.

**What kitty does now.** On Windows the terminal guard asks the console API
whether each standard handle actually names a console, rather than trusting the
C runtime's `isatty()` (which is true for the null device — CPython
[bpo-28654](https://bugs.python.org/issue28654)). The null device, a pipe, a file
and a process started without a console all now get kitty's own diagnosis and
exit 2. On Linux and macOS nothing changes: `/dev/null` was never a terminal
there, so the existing reading was already correct.

**One guard, one answer.** The fix lives in the single shared predicate
(`kitty.tui.prompts.can_interact`) that prompts and menus both consult, so the
two paths cannot drift — the same consolidation KBR-204 introduced.

## What the change does

- `can_interact()` gains a Windows branch: `GetConsoleMode` on both standard
  handles (`GetStdHandle(-10)`, `GetStdHandle(-11)`); any failure is fail-closed
  False. `STD_ERROR_HANDLE` (-12) is deliberately not consulted — the prompt
  path renders to stdout and reports to stderr, so consulting the error handle
  would add a ctypes call without adding signal.
- The decision (`nonzero GetConsoleMode return ⇒ attached`) is a one-line pure
  function (`_handle_attached`), pinned by a unit-level truth table on every CI
  leg. The ctypes call is Windows-only plumbing in its own helper
  (`_query_console_mode`), following the `bridge/manage.py` house pattern —
  `WinDLL` instantiation inside the win32-narrowed body, never at module scope.
- POSIX is untouched.
- The `can_interact`/`check_tty` docstrings carry the KBR-218 paragraph so the
  source narrates the KBR-187 → KBR-204 → KBR-218 history the design doc (§7)
  records.
- Seven test files' "simulate an interactive child" helpers extend to the new
  seam so the Windows CI leg keeps reading "interactive" when the test says so
  (the KBR-204-era helpers only patched `isatty`, which the guard no longer
  consults on Windows).

## How the change is proven

The ticket was **suspected, not reproduced** — no Windows machine was available
when it was filed. Per the ticket's own epistemics, the proof is the Windows CI
leg:

1. **Tests first** (commit 1): `tests/cli/test_stream_encoding.py` spawns
   `kitty egress` with `stdin=DEVNULL, stdout=DEVNULL` and asserts the refusal
   signature (exit 2, TTY diagnosis on stderr, no traceback). Green on POSIX at
   base (the guard already refuses `/dev/null`); the Windows leg went **red**
   with the predicted `NoConsoleScreenBufferError` traceback — that is the
   confirmation the ticket asked for. A premise test pins the platform asymmetry
   (child reports `isatty == (True, True)` on Windows, `(False, False)` on
   POSIX). A falsification case proves the harness detects a console-blind
   guard (exit 0, no diagnosis, no crash — the silent-nothing signature the
   exit-2 assertion exists to catch).
2. **Fix second** (commit 2): the console probe lands; the Windows leg turns
   green.

Unit tests in `tests/tui/test_prompts.py` pin the probe's decision in both
directions: the `GetConsoleMode` return-value interpretation on every leg with
raw values 0, 1 and a nonzero sentinel, and the four-cell AND-composition truth
table over `(stdin, stdout)` with `sys.platform` forced to `win32` — without the
truth table, a regression that made `can_interact()` always False on Windows
(silently refusing every real console) would ship green against the whole suite.

## Verification

- Full local fast selection: **6394 passed, 5 skipped, 34 deselected**.
- `ruff check`, `mypy src/kitty` (94 files, no issues), `lint-imports` (5/5
  contracts kept).
- `SYSTEM_DESIGN.md` §7 (new) carries the decision record: As Is / To Be /
  decisions-and-why / verification / known limits, including the adjacent
  `isatty` call sites (`cli/tmux_wrap.py:402`, `cli/launcher.py:290`,
  `tui/display.py:169`) that are deliberately left alone per the ticket's
  scope.
- `.requirements/20260914T162512Z_kbr218_windows_nul_tty_guard/REQUIREMENTS.md`
  (gitignored) carries the full requirements/AC/testing/implementation plan the
  `system-design-reviewer` subagent reviewed to convergence.

[KBR-218]: https://shelpuk.atlassian.net/browse/KBR-218?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


shelpuk:
**Addressed in d145bb4.**

The finding is correct and the suggested shape is what landed. The initial AC-7 wording claimed False-only patches were safe because "the real probe on Windows CI also says False" — true for CI, false for a developer's Windows machine with a real console attached (`--capture=sys` or `-s`), where `GetStdHandle` returns the console handle, `GetConsoleMode` returns nonzero, and the AND says "interactive", so the `pytest.raises(NonTTYError)` / exit-2 assertion fails. Deterministic on CI, red on a dev box — the environment-dependent shape the repo does not accept.

**Extended (nine sites, `patch("kitty.tui.prompts._handle_attached", return_value=False, create=True)`):**
- `tests/cli/test_auth_cmd.py` — the `run_oauth_for_provider` and `run_auth_openai` non-TTY tests (the two the comment names).
- `tests/test_cli_main.py::TestNonTTYExit` — the four `stdin.isatty=False` sites (`test_interactive_command_exits_2_with_a_message`, `test_hijacked_setup_explains_the_redirection`, the `test_no_redirection_note_when_nobody_was_redirected` parametrize, and `test_non_tty_error_is_still_raised_by_check_tty`).
- `tests/tui/test_profile_menu.py::TestRunProfileMenuGuard::test_non_tty_raises`
- `tests/tui/test_setup_wizard.py::TestRunSetupWizard::test_non_tty_raises`
- `tests/tui/test_egress_menu.py::TestMenuShape::test_non_tty_is_rejected` (the comment's truncation cut these last three; same gap, same fix).

**Also updated** `SYSTEM_DESIGN.md` §7.4 case 5 / §7.5 to record the seam rule in both directions (True for proceed-expectation, False for refusal-expectation) with the dev-box rationale, and the grep invariant now reads `return_value=…` rather than `return_value=True` so the negative direction is covered by the rule too. Local gates green after the change.

shelpuk:
**Addressed in 8557b4b.**

The finding is correct and the suggested shape is what landed. The four parameterised `_mock_tty` helpers now use `side_effect=cycle([is_tty, out])` instead of `side_effect=iter([is_tty, out])`:

- `tests/tui/test_prompts.py::_mock_tty`
- `tests/tui/test_menu.py::_mock_tty`
- `tests/tui/test_setup_wizard.py::_mock_tty`
- `tests/cli/test_auth_cmd.py::_mock_tty`

The helper comment in each file now says why `cycle`: a wizard or OAuth flow invokes `can_interact` several times inside one patched block (the top-level `check_tty` plus one per prompt and per `SelectionMenu.show()`), and an iterator exhausts on the third call with `StopIteration` on the Windows CI leg — a POSIX leg never sees it because `can_interact` reads `isatty` there.

The truth-table unit test (`test_can_interact_and_composes_the_console_probe`) is unchanged and intentionally stays on `iter` — it patches `_query_console_mode` (the plumbing), not `_handle_attached`, and each parametrised cell calls `can_interact()` exactly once, consuming exactly the two iterator values. Same for the inline `return_value=True` / `return_value=False` sites.

`SYSTEM_DESIGN.md` §7.4 case 4 now names the correct patch target (`_query_console_mode`) and `REQUIREMENTS.md` R7 records the `cycle`-not-`iter` rationale as a decision, so the next reader does not re-derive it.

Local gates green after the change.

PR #148, opened 2026-09-14
KBR-34: Chat Completions reader (T-A2) — closes G37, proves merge rule

shelpuk:
## Context

For Product Owners — what this PR ships, in product terms

**The Chat Completions reader.** The bridge speaks Chat Completions on every non-native route — when Claude Code's messages are translated for OpenAI, Ollama Cloud, or any provider speaking OpenAI-compatible. Today four of the seven wire projections exist (Anthropic Messages, Gemini, Responses — three readers), but the Chat Completions reader is the one the bridge compares against Claude Code's body on the majority of real traffic. Without it, every Messages↔Chat Completions comparison is a phantom delta, and §3.3.1b's merge rule has no proof it ships the same shape across readers.

**What this PR lands:**

- The Chat Completions reader (`tests/harness/reader_chat_completions.py`) — projects a CC body into the wire-independent form, importing nothing from `src/kitty/bridge` (§3.3.1's independent-oracle rule). Written against `openai/openai-openapi` master, retrieved 2026-09-14, the same source the design doc cites for G37.
- The **convergence proof** (KBR-25's scope addition): the CC and Anthropic Messages encodings of one worked tool exchange project to identical `Conversation` values, indices included. T-A2's claim that §3.3.1b's merge rule is shared, not just stated.
- The **G37 closure** (KBR-199): both cache-breakpoint spellings (`cache_control` on OpenRouter's CC dialect; `prompt_cache_breakpoint` on OpenAI's GPT-5.6+ dialect) fill `Part.cache_control` verbatim — the slot is `Mapping[str, Any] | None` and §3.3.1's "carried whole, not reduced" rule applies to a spelling with no TTL the same way it applies to one with.
- The first reader that **reads** the parallel-tool-use canonical address (`envelope.extra["parallel_tool_calls"]`) directly — the wire key is the canonical spelling; KBR-205 wrote the Anthropic reader's side, this reader reads the same address.

**What this PR doesn't change for users (no behaviour change):**

The CC reader projects everything the schema publishes into the wire-independent form. The oracle can now compare the Messages and Chat Completions encodings of the same conversation and report a delta only when one actually happened.

## For reviewers

**Files changed:**

- `tests/harness/reader_chat_completions.py` (new, ~830 LOC)
- `tests/harness/test_reader_chat_completions.py` (new, 49 tests)
- `tests/harness/contract.py` (cherry-pick from KBR-205: `IGNORED_BLOCK_FIELDS`, `PARALLEL_TOOL_CALLS_KEY`, `ToolDecl.type`)
- `tests/harness/reader_anthropic_messages.py` (cherry-pick from KBR-205: parallel-tool-use mapping, declared-ignored handling, tool `type`)
- `tests/harness/test_reader_anthropic_messages.py` (cherry-pick: 21 new tests)
- `.system_design/TEST_SUITE.md` (cherry-pick + KBR-34 additions: §3.3.1b parallel-knob rule, §7.4.1 CC reader section, §9.2 G37 CLOSED, §11 Q15 scope note + Q16)

The KBR-205 cherry-pick is the dependency KBR-34 explicitly named in the ticket — `PARALLEL_TOOL_CALLS_KEY` and `IGNORED_BLOCK_FIELDS` are the vocabulary this reader uses. Both KBRs are independently reviewable, and they are stacked in one PR because the CC reader cannot function without KBR-205's vocabulary.

**Coverage of the four-failure modes the system-design-reviewer named:**

1. The six unclassified CC control fields (`audio`, `moderation`, `functions`, `function_call` deprecated) — all four ride at their wire keys now; checked against `openai-openapi` master's actual `CreateChatCompletionRequest` property list.
2. The fourth `tool_choice` shape (`allowed_tools`) — `mode` follows the same string-to-canonical mapping as the top-level strings; an unknown mode raises `UnreadableBodyError`.
3. Refusal — both spellings (content part and message-level field) project as `Opaque(kind="refusal", digest=text_digest(...))`, matching T-A3.
4. The convergence pin — the interleaved exchange does not converge across the two readers, and that is the design (§3.3.1b: clause 3 is vacuous on CC, active on Messages). The test documents the asymmetry; the convergence claim is pinned on the simple exchange that does converge.

**Verification:**

- `pytest tests/harness/` — 1551 passed, 13.36s.
- `ruff check` clean.
- `mypy` clean on the harness modules.
- `lint-imports` 5/5 contracts kept.

## Task requirements

- `.requirements/20260914T195500Z_kbr_34_chat_completions_reader/REQUIREMENTS.md`


shelpuk:
## Round 3 — all 9 findings applied (`967a541`)

Thank you for the careful round-3 review. Every finding was real. Applied in commit `967a541`:

| # | Finding | Action |
|---|---------|--------|
| W1 | `store` and `user` had no direct test | Added to R1.6a's parametrise list; R1.6a now covers all 12 of `_PUBLISHED_EXTRA_KEYS` |
| W2 | §11 Q16 cited a co-occurrence case no test asserted — and the reader was wrong too: the second spelling on a co-occurrence was **silently dropped**, not residualised | Two fixes: (a) `_read_cache_control` now residualises the second spelling at its own path (silent drop is the shape M16 and G37 exist to prevent); (b) new **R6.4** test pins first-fills / second-residuals on a part carrying both spellings |
| S1 | Reader docstring claimed `functions`/`function_call` residualise; they ride at `extra[<wire key>]` | Docstring fixed; names the row-plan owed to M16's twin |
| S2 | `_SAMPLING_KEYS` "is the identity" overstates — it excludes `top_k` and CC never carries `stop_sequences` | Docstring now says thirteen of fifteen are identity, names the rename and the exclusion |
| S3 | `TestMergeRule` numbering R4.1/R4.3/R4.2 didn't match definition order | Renumbered to R4.1/R4.2/R4.3 |
| S4 | `test_a_wrongly_typed_content_part_type_raises` named one defect, exercised another | Renamed to `test_a_user_role_with_a_refusal_part_type_is_unreadable`, docstring updated |
| S5 | R6.3 docstring said "at either spelling" but tested one | Docstring names `cache_control` directly and points at the re-spelled-sibling pin; the co-occurrence test (R6.4) is the stronger coverage |
| S6 | CC's `parallel_tool_calls` was unconditionally added to `consumed`; the Anthropic reader's analogous site is conditional on `bool` | **Code change**: `isinstance(raw, bool)` now gates both `extra[parallel_tool_calls]` and `consumed`; a wrongly-typed value residualises at the bare name and the run fails closed. Cross-reader behaviour now matches |
| S7 | R1.6a docstring said "six unclassified fields" but listed four | Names the four B1 keys plus the eight earlier ones, with the right count |

**Verification** — `pytest tests/harness/` 1565 passed (13.64s), ruff clean, mypy clean, import-linter 5/5 kept. CI re-runs on `967a541`.

One substantive note on W2: the doc's claim was also backwards from what the code did. The doc said "the second fills the slot and the first residualises" — the code silently dropped the second (neither fills nor residualises). The fix preserves the doc's *intent* (the second is named, not swallowed) while correcting the ordering (first fills, second residualises, per `_CACHE_KEYS` order).

shelpuk:
## Round 4 — all 6 warnings + 2 suggestions applied (`2c2b448`)

The round-4 review found that round 3 introduced three defects of its own, surfaced two pre-existing silent-drop holes, and that one design-doc update didn't ship with its code change. All eight findings were real. Applied in `2c2b448`:

| # | Finding | Action |
|---|---------|--------|
| W1 | `parallel_tool_calls: null` silently dropped after the round-3 conditional refactor | Fixed: `None` now consumes the key (totality) without writing `extra[parallel_tool_calls]` — the `cache_control` precedent (the wire key carries no instruction, the bridge does not invent one). New **R1.5b** test pins the behaviour |
| W2 | R6.4 was hash-seed-dependent — `_CACHE_KEYS` was a `frozenset` with non-deterministic iteration order, so the "first non-null wins" rule would flip between runs | **Code change**: `_CACHE_KEYS` is now an ordered `tuple`, so the rule is stable across Python's hash randomisation. New **R6.5** test runs the projection twice and confirms the same shape both times |
| W3 | Assistant message's `audio` and `function_call` silently dropped — claimed as mapped, never extracted | Fixed: removed from the mapped set; they now residualise at the message's path, the run fails closed, and the residual entry is the hook a future reader that grows an `Audio` part or a `function_call` mapping consults. New **R3.7b** test |
| W4 | R6.3 docstring pointed at a test that exists only in the Anthropic reader, not in the CC reader's declared-ignored suite | Docstring now points at R6.4 — the co-occurrence pin, which is the actual coverage |
| W5 | §11 Q16 still cited the pre-round-3 test name and the wrong co-occurrence direction | Design doc now names R6.4 and states the correct ordering ("first non-null wins, the other residualises" per `_CACHE_KEYS` order) |
| W6 | `cache_control` on an undecodable-base64 image part silently lost | Fixed: `cache_control` residualises at its own path when the decode fails, so a body carrying a undecodable image with a breakpoint is one the bridge cannot faithfully forward. New **R6.3b** test |
| S1 | R1.6a test did not cover a wrongly-typed value for any `_PUBLISHED_EXTRA_KEYS` member | R1.6b restored and its docstring sharpened: the value is carried whole, not type-checked (§3.3.1a's compared-whole rule) |
| S2 | `tool_choice.allowed_tools.tools` list not pinned as invisible to the oracle | **Code change**: the list residualises at `tool_choice.allowed_tools.tools` so the body is accounted for. New **R2.4b** test |

**Verification** — `pytest tests/harness/` 1570 passed (12.16s, up from 1565 — 5 new tests), ruff clean, mypy clean, import-linter 5/5 kept. CI re-runs on `2c2b448`.

Three of the eight findings changed code behaviour (W1, W2, W6); two more changed code in ways the reviewer flagged (W3, S2); the remaining three were doc/docstring/test-rename fixes. CI re-validates from scratch on the new commit.

shelpuk:
## Round 5 — 1 warning + 2 suggestions applied (`f11eae9`)

The round-5 review caught a real hole in the round-4 fix (the parallel non-base64 data-URL branch still silently dropped `cache_control`) plus two test-coverage gaps. All three applied:

| # | Finding | Action |
|---|---------|--------|
| W1 | `cache_control` on a non-base64 data URL image part silently dropped (round-4 fix missed the `_ANY_DATA_URL` parallel branch) | **Code change**: the `_ANY_DATA_URL` branch now residualises `cache_control` at its own path — the same shape the base64-decode-failure branch got in round 4. A silent drop is the M16/G37 shape the residual rule exists to prevent, on both branches now |
| S1 | R1.6a parametrise still does not cover wrongly-typed values for 11 of 13 `_PUBLISHED_EXTRA_KEYS` members | Added **R1.6c** — a parametrised test pinning a wrongly-typed value at each of the 13 published-extra keys carries whole (§3.3.1a's compared-whole rule), matching R1.6b's `audio` case across the full set |
| S2 | R6.5 determinism pin runs in a single process, weaker than its docstring's cross-PYTHONHASHSEED claim | Rewritten to run the reader under **three different `PYTHONHASHSEED` values (0, 1, 2)** via subprocess, asserting the same projection shape in each. A within-process assertion cannot reach across the runtime's hash seed, so this is the only pin that actually proves the claim |

**Verification** — `pytest tests/harness/` 1583 passed (12.69s, up from 1570 — 13 new R1.6c cases), ruff clean, mypy clean, import-linter 5/5 kept. CI re-runs on `f11eae9`.

One reviewer note worth recording: the round-5 review's claim that round-4's commit message promised parametrised wrongly-typed coverage for all 12 keys was correct — the commit said "R1.6b restored and its docstring sharpened" but the docstring change did not add coverage. Round 5's fix actually adds it. The lesson: a commit message's coverage claim should be backed by the diff, not by intent.

shelpuk:
## Round 6 — 2 warnings + 1 suggestion applied (`cca6b84`)

The round-6 review caught a real design-rule violation in the round-5 state — the CC reader was writing `parallel_tool_calls: True`, the CC default, contradicting §3.3.1b's "write only non-default" rule and the Anthropic reader's conditional-write pattern. All three findings applied:

| # | Finding | Action |
|---|---------|--------|
| W1 | CC reader writes `parallel_tool_calls: True` (the CC default); §3.3.1b says only non-default | **Code change**: `True` is now consumed but not written to `extra[parallel_tool_calls]` — the absence is the canonical form, and writing the default would invent a second field some providers reject. `False` is the non-default case the oracle names (R1.4). New **R1.5c** test pins that `True` produces no entry |
| W2 | R1.6c parametrise covered 13 of 14 `_PUBLISHED_EXTRA_KEYS` members; `functions` was missing and the docstring said 13 keys | `functions` added to the parametrise; docstring names all 14 (including the two deprecated top-level spellings) |
| S1 | R1.6c docstring claimed the pin value is the wrong type for every key; incorrect for the five string-typed keys (`metadata`, `service_tier`, `verbosity`, `user`, `reasoning_effort`) | Docstring rewritten to name both shapes: the string is the wrong type for the dict-shaped keys and the right type for the string-shaped keys; the pin holds either way because the reader does not enforce a type contract it was never asked to enforce |

**Verification** — `pytest tests/harness/` 1585 passed (12.90s, up from 1583 — 2 new test cases), ruff clean, mypy clean, import-linter 5/5 kept. CI re-runs on `cca6b84`.

W1 was a design-rule defect, not a test-coverage one — the `True` write would have produced a false I1 delta on every CC inbound body whose upstream projection was empty (the cross-format case the design rule exists to prevent). Round 5's PR-comment-side claim that "True is the CC default so a body carrying false is a non-default delta" was already in the docstring but the implementation disagreed; round 6 caught the disagreement.

shelpuk:
## Round 7 — no findings; two docstring items in reviewer's prose addressed (`50b4a02`)

The round-7 review reported **"No findings"** and explicitly withheld 2 minor doc polish items per its own rule that suggestion-level findings are scoped at round 6. Both items the reviewer named in its prose were addressed in `50b4a02`:

| # | Reviewer prose | Action |
|---|---|---|
| 1 | R1.6c docstring self-contradicted "wrong type for the nine dict-shaped keys" with a parenthetical listing all 14 keys | Rewritten: the pin value is a string regardless of the schema's right type; the test pins the carried-whole behaviour either way (§3.3.1a's compared-whole rule) |
| 2 | R1.6a docstring said "four B1 keys (`audio`, `moderation`, `functions`, `function_call`, all deprecated)" — only `functions` and `function_call` are deprecated | Rewritten: 14 keys total = 10 current + 2 current-but-peripheral (`audio`, `moderation`) + 2 deprecated top-level spellings (`functions`, `function_call`); each group named |

**Verification** — `pytest tests/harness/` 1585 passed (no new tests; the docstring fixes don't change behaviour), ruff clean.

The reviewer at the actionable level (warnings, criticals) has stopped providing useful feedback: round 7 was empty, and the reviewer explicitly scoped its remaining 2 items out of formal findings. Per the loop's contract, the review cycle is closed. PR #148 is ready to merge from the CI side once the gate finishes on `50b4a02`.

shelpuk:
## Round 8 — 1 warning + 3 prose items applied (`e898846`)

Round 8 surfaced a real cross-reader asymmetry the prior rounds did not exercise, plus three doc/naming items. All four applied:

| # | Finding | Action |
|---|---------|--------|
| W1 | Anthropic reader residualised `disable_parallel_tool_use: null` while CC reader consumed `parallel_tool_calls: null` — six readers cannot quietly disagree about what a null flag means | **Code change**: the Anthropic reader's `_read_tool_choice` now treats `flag is None` as absent-by-null (the `cache_control` precedent) and consumes the key for totality, matching the CC branch. New test `test_disable_parallel_tool_use_null_is_treated_as_absent` pins the Anthropic side; R1.5b already pins the CC side |
| P1 | R7.2's `wrong_merged`/`right_merged` variable names were swapped relative to their content (`wrong_merged` was `[ToolResult, Text]`, which is the right order per the merge rule) | Renamed to `correct_merged`/`re_sorted` matching the content |
| P2 | R1.6a docstring said "fourteen keys" while the parametrise covers twelve | Docstring now says "twelve of fourteen" and names the two R1.6 covers separately (`metadata`, `service_tier`); the union of R1.6 + R1.6a is the full set |
| P3 | The non-base64 data URL + cache_control case shares a code path with the base64-decode-failure case but was not directly tested | Added **R6.3c**, pinning the `_ANY_DATA_URL` branch residualises both URL and `cache_control`, mirroring R6.3b |

**Verification** — `pytest tests/harness/` 1586 passed (11.74s, up from 1585 — 2 new tests), ruff clean, mypy clean, import-linter 5/5 kept. CI re-runs on `e898846`.

W1 is the substantive finding: the design rule §3.3.1b is that six readers cannot quietly disagree, and a corpus entry that fires `disable_parallel_tool_use: null` on the Anthropic wire would have failed closed while the same intent on the CC wire would have projected green. The fix is one branch (the `if isinstance(flag, bool)` becomes `if flag is None or isinstance(flag, bool)`) plus the cache_control precedent's "absent-by-null" pattern. Round 8 caught what the prior rounds did not exercise; the lesson is that cross-reader asymmetry belongs on a check-list the test sweeps every reader against.

shelpuk:
## Round 9 — 2 warnings, both silent-drop holes closed (`5c1bae0`)

Round 9 surfaced two pre-existing silent-drop holes that contradict the reader's own module-level totality claim and §3.3.1's "fail closed at every depth" rule. Both closed:

| # | Finding | Action |
|---|---------|--------|
| W1 | `_system_parts` silently dropped `cache_control` on system/developer content parts and silently dropped every other key the grammar does not model on those parts | **Code change**: `_read_system_messages` and `_system_parts` now take a `residual` parameter and thread it through. `_system_parts` reads `cache_control` onto the Text part via `_read_cache_control` (the same recipe user-content-part code uses) and residualises every other key via `_residualise(part, {"type", "text", *_CACHE_KEYS}, ...)`. New **R3.2b** test pins `cache_control` on a system content part |
| W2 | `_read_tool_choice` silently dropped every sub-key of `tool_choice` and of its member other than those explicitly read | **Code change**: after each branch extracts the needed keys (type + member.name for the named form; type + allowed_tools.mode for the restricted set), every other key on the tool_choice object and its member residuals at its own path. Two new tests pin both branches: `unknown_keys_on_a_tool_choice_residualise` and `unknown_keys_on_an_allowed_tools_member_residualise` |

**Verification** — `pytest tests/harness/` 1589 passed (12.51s, up from 1586 — 3 new tests), ruff clean, mypy clean, import-linter 5/5 kept. CI re-runs on `5c1bae0`.

Both findings were the same shape: code paths that consumed keys without reading them and without residualising them, violating the reader's own "every key of the body, at every depth, is either mapped or placed in residual" claim. The reader's docstring is now honest — system parts and tool_choice members both honour the totality rule.

shelpuk:
## Round 10 — review loop closed; two minor prose items addressed (`7c09d75`)

The round-10 review reported **"No findings"** at the warning/critical level. The reviewer explicitly withheld two minor prose items, both now addressed in `7c09d75`:

| # | Reviewer prose | Action |
|---|---|---|
| 1 | CC reader's `_residualise` kept `mapped: set[str]` while the Anthropic reader's was widened to `Collection[str]` | **Signature aligned**: CC's `_residualise` now also accepts `Collection[str]`, docstring mirrors the Anthropic wording ("accepts a `set` or a `frozenset` — the per-block call sites build a fresh `set` from a literal, while `_TOOL_KEYS` is a `frozenset` so a reader cannot quietly grow it mid-request") |
| 2 | Two R3.x wildcard test names in `TestMessages` didn't match the surrounding numbering | Renamed to R3.14 and R3.15 |

**Verification** — `pytest tests/harness/test_reader_chat_completions.py` 87 passed, ruff clean, mypy clean.

## Review loop summary

Ten rounds of bot review, and the reviewer has stopped providing useful feedback at the actionable level. Every round's findings applied:

| Round | Findings | Commit |
|---|---|---|
| 1 | 5 blockers (design reviewer) + C1 vacuous pin | `eb28236` |
| 2 | 2 warnings (B1 coverage, merge-rule pin) | `8db8474` |
| 3 | 9 findings (B1–B5 + 4 C) | `967a541` |
| 4 | 6 warnings + 2 suggestions (silent drops, hash seed, cache_control on undecodable image) | `2c2b448` |
| 5 | 1 warning + 2 suggestions (the `_ANY_DATA_URL` branch, R1.6c coverage, cross-`PYTHONHASHSEED`) | `f11eae9` |
| 6 | 2 warnings + 1 suggestion (`parallel_tool_calls: True` design-rule defect, R1.6c missing `functions`, docstring) | `cca6b84` |
| 7 | No findings; 2 prose items addressed | `50b4a02` |
| 8 | 1 warning + 3 prose items (cross-reader null-flag asymmetry, R7.2 names, R1.6a count, R6.3c branch) | `e898846` |
| 9 | 2 warnings (`_system_parts` silent drop, `_read_tool_choice` silent drop) | `5c1bae0` |
| 10 | No findings; 2 prose items addressed | `7c09d75` |

The substantive defects all closed: every silent-drop hole the reviewer surfaced is closed; the design-rule violation (`parallel_tool_calls: True` write) is fixed; the cross-reader asymmetry (Anthropic vs CC on null flags) is aligned; the convergence pin is non-vacuous and pinned through the reader; `_CACHE_KEYS` is deterministic across `PYTHONHASHSEED`.

PR #148 is ready from the agent side. CI re-runs on `7c09d75`; per the owner's instruction the PR is not merged by the agent.

PR #150, opened 2026-09-14
T-B4 (KBR-43): scripted failure library for the bridge recorder

shelpuk:

The test harness that proves kitty-bridge's promises could, until now, only simulate a *well-behaved* provider. Every ticket that needs to prove how the bridge behaves when things go wrong — a provider firewall blocking the account, a context-window rejection, a stream that dies mid-sentence — had to hand-build its own failure, and each hand-built failure is a chance for a test to quietly test the wrong thing.

This change gives the harness a **scripted failure library**: one reviewed place that knows how to make an upstream misbehave in every way the bridge recognises, and to cut a live stream at exactly the four points our design names (before the first byte, after text reaches the client, mid tool-call arguments, and after content but before the closing event). Everything is forced by scripted byte sequences, never by timing, so the failures reproduce identically on every platform CI runs.

Seven planned tickets — the streaming-recovery, failover, agent-smoke, and load work (T-D4, T-G7, T-I5, T-I7, T-I8, T-I11, T-K4) — build directly on this. No product code changes; this is test infrastructure only.

## What ships

**`tests/harness/failures.py`** — the library. Builders, each returning a `Responder` closure for T-W4's recorder seam:

- `error_status(fmt, status, ...)` — protocol-native error envelopes for both served formats (Anthropic Messages / Chat Completions).
- `cloudflare_block(fmt)` — a 403 whose body trips `is_cloudflare_block` (the substring is hardcoded here, not imported from kitty — the library shares no constant with the product).
- `context_too_large(fmt, status=400|413)` — a rejection the bridge's `_is_context_too_large_error` classifies as oversized.
- `empty_response(fmt, stream=)` — the contentless success shapes. Recorded caveat: on the Chat Completions wire a role chunk is a *well-formed skeleton* (KBR-232), not an empty-retry; on Messages the preamble hold judges it.
- `drop_at(fmt, point)` + `InjectionPoint` — the four §6.3.1 mid-stream disconnects: `BEFORE_FIRST_BYTE` (pre-emission), `AFTER_TEXT`, `MID_TOOL_ARGUMENTS` (byte-exact partial `'{"arg1":"v'`, tool block deliberately left open), `BEFORE_TERMINAL` (everything but `message_stop` / `[DONE]`). Uses `Reply.abort()` (KBR-189: flushes queued bytes before dropping).
- `scripted(*replies)` — per-attempt dispatch; raises `ScriptExhausted` on an unexpected extra attempt instead of silently falling back to a success.
- `success(fmt, stream=)` — the recorder's minimal shapes, so a failover script can end clean.
- `sse_frame` / `cc_chunk` / `CC_DONE` — one frame builder per wire grammar (Anthropic carries `event:` lines; CC is `data:`-only).

**`tests/harness/test_failures.py`** — 33 tests: every builder driven through the real `RecordingUpstream`; the bridge's own detectors (`is_cloudflare_block`, `_is_context_too_large_error`) imported by the *tests* as oracles; and the falsification the plan's harness rule requires:

> A hand-rolled `drop_at` variant that calls `transport.abort()` — the KBR-189 defect — loses a queued 16 MiB overrun where `Reply.abort()` delivers every byte, with a vacuous-guard on the queue size and §7.2's barrier rule recorded beside the assertion.

A structural guard pins the library's independence: `failures.py` must not import `src/kitty` (verified to fail on an injected defect).

## Review trail

- Blocker KBR-27 (T-W4, the recorder) merged via PR #60 — verified on `origin/main` before starting.
- `system-design-reviewer` pass on the requirements found two blockers (falsification untestable per-leg in its first form; a kitty-import contradiction) and eight concerns — all applied before implementation (recorded in the task's `REQUIREMENTS.md` §7).
- `code-reviewer` pass on the implementation found four test-coverage gaps (missing AC assertions, two CC branches untested) — all fixed before this PR.

## Definition of done (plan §1.3)

- [x] Local: `pytest -m "l1 or l2"` (33 new tests), `ruff check .`, `mypy src/kitty`, `lint-imports` — all green.
- [x] Lands on `main` alone; no product (`src/`) changes; layer markers unchanged (harness tests default `l1`).
- [ ] CI on the Windows and macOS legs (watching this PR).

Design: TEST_SUITE.md §6.3.1, §7.2 · Plan: T-B4 · Jira: KBR-43


shelpuk:
Windows-leg fix pushed (b84ed9e): the falsification's broken-variant assertion was Linux/macOS-shaped. On Windows' Proactor loop every write — headers and SSE frames included — is still queued at `transport.abort()`, so the client receives nothing at all; on selector loops the frames reach the kernel and only the queued 16 MiB overrun is lost. Both outcomes prove the KBR-189 defect; the common assertion (the overrun did not survive the abort) now runs on every leg, with the platform-specific consequence asserted beside it and the reasoning recorded in the test docstring (§7.2 documents both shapes). All 33 tests pass locally; watching the re-run.

shelpuk:
Review-bot findings addressed in 18c8ac8:

**Warning** — Recorder fixture now calls `assert_all_paths_matched()` at teardown (test_recorder.py's pattern), so a wrong-format reply fails loudly instead of reading as an empty body.

**Suggestions:**
- `context_too_large`'s CC `code` field is now `str(status)`, consistent with `error_status`'s CC envelope and the documented CC contract.
- `_check_fmt` builds its error message inside the if, after an `isinstance(fmt, WireFormat)` guard, so a non-enum caller sees a `ValueError` rather than an `AttributeError` from the format-string interpolation.
- The lock-step docstring now states the honest scope: the test catches the abort semantics, not the wire grammar (a `frames_for` regression that affected both sides equally would pass this test).

33 tests pass locally; watching the CI re-run.

PR #152, opened 2026-09-14
KBR-41: curl_cffi recorder + harness TLS, and the OAuth refresh leg

shelpuk:

Today the harness can see four of the bridge's five upstream client configurations, but the **fifth is invisible**: every conversation with an OpenAI ChatGPT subscription account rides `curl_cffi` with a Chrome TLS fingerprint, and the harness's plain-HTTP recorder cannot look at a stream that has not finished its handshake. Two register rows the product's correctness depends on — P13 (fourteen Chat-Completions parameters dropped by `_cc_to_responses`) and P17 (unconditional `stream: true` / `store: false`) — were therefore untestable, and the OAuth refresh leg (which KBR-161 moved onto this stack) had no observation point at all. T-B2 closes that gap.

This PR adds a **TLS-terminating local recorder**, a **`curl_cffi` transport** for the bridge fixture, and the two seams that point the product at the recorder. After it lands:

* every test that drives a real bridge can now observe OpenAI subscription traffic;
* the OAuth refresh leg is captured, with credentials masked before they reach an assertion diff or a CI log;
* a TLS-terminating recorder logs every accepted connection at the moment of accept, not after the handshake, so a failed TLS negotiation no longer hides a connection from the containment harness.

## What changed

**New modules**

- `tests/harness/curl_recorder.py` — `CurlRecordingUpstream` (TLS-terminating, serves `OPENAI_RESPONSES` on `*/responses` and form-encoded token grants on `*/oauth/token`), `CurlLoggingServer` (the protocol-factory override that logs every accept before the TLS handshake begins).
- `tests/harness/curl_cffi.py` — `CurlCffiTransport` (the registry entry), the `redact_oauth_form_body` mask, and the two seam context managers (`codex_backend_url`, `oauth_refresh_endpoint`).
- `tests/harness/test_curl_recorder.py` — the 14 conformance checks over TLS, vocabulary, socket-level logging with two falsification directions (missing-record and double-record), and TLS termination.
- `tests/harness/test_curl_cffi.py` — interface, redaction, seams, end-to-end refresh-leg through a stale-token session, bridge round-trip.

**Modifications**

- `tests/harness/bridge.py` — `BridgeFixture` and `assert_transport_reaches_its_recorder` gain a `key=` parameter; the existing two transports default to the previous behaviour, the curl_cffi row carries the OAuth session file through a setup callable. `transport()` forwards `**kwargs` to the factory so a transport can request its own TLS configuration.
- `tests/harness/test_bridge.py` — `CONFORMANCE_CASES` gains the curl_cffi row; the meta-test drives it; the registry-completeness test asserts the extended set.
- `src/kitty/auth/oauth_session.py` — module docstring corrected (KBR-212): the refresh leg runs on a second session built by the same builder, not "the same session".
- `.system_design/TEST_SUITE.md` — §7.2 table updated; §7.2.3 records what this task settled, naming the socket-level seam, the two constant-swap seams, and the per-transport redaction (and why the per-transport default lives inline rather than on `UpstreamTransport`).

## Decisions

Three product-affecting decisions, all confirmed by the owner on 2026-09-14:

1. **§7.2.1's connection-logging limitation → observe at socket level.** The accept-time append is in `CurlLoggingServer.__call__`, which asyncio calls before the SSL transport is built. The peer port is filled in by `connection_made` once the handshake completes; a record that keeps `-1` is the failed-handshake shape. Mutation-verified: dropping the accept-time append fails the failed-handshake test; restoring a second append fails the double-record test.
2. **Body redaction → per-transport on `captures`, not on the `UpstreamTransport` Protocol.** `CapturedRequest.body` stays raw (T-C6's malformed body must reach a reader untouched); `CurlCffiTransport.captures` returns masked copies. **Not** a `redact_body` method on the `UpstreamTransport` Protocol — the Protocol is `@runtime_checkable` and three suites assert `isinstance(..., UpstreamTransport)`, so a Protocol member must be implemented by every instance, which would defeat the "inherit the no-op" property.
3. **KBR-211 (`_reset_curl_session` orphans) → deferred to its own ticket.** Fixing it means closing under a live request — the shape curl_cffi upstream issue #675 reported as a segfault. It needs the measurement this recorder enables and should be reviewed as its own change. KBR-212 (one-sentence docstring fix) is in scope and landed here.

## Verification

- **ruff**: clean.
- **mypy src/kitty**: clean.
- **import-linter**: all five contracts kept.
- **pytest tests/harness/** (1472 tests): all green.
- **pytest tests/test_curl_cffi_transport_contract.py**: all green.
- **pytest tests/harness/test_register_agreement.py** (register ↔ design doc ↔ source): all green.
- **pytest tests/harness/test_bridge_falsification.py** (the four conformance defects, T-W8's R6): all green.
- **pytest tests/test_layer_markers.py / tests/test_layer_selection.py** (the two layer guards): all green.
- Full l1 suite (post-merge) running; results appended to this PR when available.


shelpuk:
### Re: Claude Code review (review id 5202419895)

All six findings addressed in commit \`88c5fcd\`:

**Warnings**

1. **bind() docstring's false RuntimeError** — removed; \`bind()\` no longer reads \`base_url\`, so the precondition no longer holds.
2. **start() cleanup can itself raise on a half-entered seam pair** — restructured: the two seams now enter in a nested try so a failure entering the second rolls the first back, and only a fully-entered pair is published on the transport. \`stop()\` then always exits seams this transport owns, never a fresh context manager.
3. **asyncio.sleep(0.05) condition wait** — replaced with a \`_wait_until\` helper that polls on a bounded timeout (default 2.0 s, 5 ms between polls). Both socket-level tests use it.
4. **stop() unconditionally unsets CODEX_CA_CERTIFICATE** — bound off: \`bind()\` records the previous value in \`_env_state\`; \`stop()\` restores it (or pops if it was unset). A transport that never called \`bind()\` does not touch the environment. The pre-bind value is recorded only on the first bind — a second \`bind()\` would otherwise overwrite it with the first bind's own value and lose the original. (This second-bind bug surfaced as \`test_session_created_with_impersonate\` failing when \`test_bind_returns_one_adapter_for_the_transport_lifetime\` ran before it; both pass together now.)

**Suggestions**

1. **Duplicate \`_seed_oauth_session\`** — moved to \`harness.curl_cffi.py\` as \`seed_oauth_session(tmp_path, *, stale=False)\`; both \`test_bridge.py\` and \`test_curl_cffi.py\` now import it.
2. **Three \`TestThroughARealBridge\` tests duplicate transport construction** — extracted to a local \`_build_transport\` helper.

**Verification**

- \`ruff\`, \`mypy src/kitty\`, \`lint-imports\` all green.
- Harness suite: 414 tests pass with all curl_cffi + provider_aiohttp + bridge tests.
- Cross-suite: \`tests/harness/test_curl_cffi.py\` runs cleanly before \`tests/providers/test_openai_subscription.py::TestTlsImpersonation::test_session_created_with_impersonate\` (the leak the second-bind bug caused is gone).
- Mutation-verified: reverting to the pre-fix \`bind()\` triggers \`test_session_created_with_impersonate\` failure with \`'verify=/tmp/.../kbr41-ca.pem'\` in the \`AsyncSession\` call.

shelpuk:
### Re: follow-up suggestion — same \`bind()\` docstring defect in AiohttpTransport / ProviderAiohttpTransport

Verified the claim against the code at HEAD — the suggestion is **inaccurate**, and I have not made the change.

- \`tests/harness/bridge.py:444\` is \`AiohttpTransport.bind()\`. Its body is a single line: \`return _ADAPTER_FOR_FORMAT[self.format](), {"base_url": self._recorder.base_url}\`. \`self._recorder.base_url\` (T-W4's \`RecordingUpstream.base_url\` property) **does** raise \`RuntimeError("recorder is not running; call start() first")\`. The \`Raises: RuntimeError\` block accurately describes this.
- \`tests/harness/provider_aiohttp.py:191\` is \`ProviderAiohttpTransport.bind()\`. Its body starts with \`base_url = self._recorder.base_url\` — same property, same \`RuntimeError\` path. \`Raises:\` block is accurate.

Only \`CurlCffiTransport.bind()\` had the docstring defect: when I removed the \`base_url\` read (the constant-swap seam makes the URL irrelevant here), the \`Raises:\` block no longer matched the implementation. That was addressed in commit \`88c5fcd\`.

Leaving the other two docstrings as written.

PR #156, opened 2026-09-14
KBR-46: corpus truncation and compaction threshold pairs (T-C3)

shelpuk:
## Context (for Product Owner)

The kitty-bridge test suite is building a "golden corpus" — recorded
Claude Code conversations used to prove the product never silently alters
what a coding agent sends. This ticket contributes the four hardest
fixtures in that corpus: the pairs that sit exactly on two size boundaries
where the bridge mutates the request before forwarding it to the provider:

- **`tool_result_under_limit`** — one tool result of exactly 50 000 chars,
  the largest size that does NOT trigger the bridge's tool-result
  truncation. The strict `>` comparison in three truncation sites
  (`server.py:7269`, `:7286`, `:7320`) is what pins this number.
- **`tool_result_over_limit`** — 50 001 chars, the smallest that does.
- **`compaction_budget_under`** — a transcript sized so the
  bridge-measured length is exactly 2 800 000 chars, the largest
  short-circuit size on the static fallback budget. M4/M5 complement.
- **`compaction_budget_over`** — filler alone crosses 2 800 001 chars and
  carries one oversized tool result so M4's second condition is also
  present; M5 still fires after M3 truncates the result.

The 50 000 / 50 001 boundary is the one a code reviewer can read off
three truncation sites and confirm. The 2 800 000 boundary is profile-
dependent in the runtime (the bridge's budget derives from the model's
context window), so this commit also pins the calibration honestly: the
entries are tuned for the static fallback path, and the manifest
documents that 1M-token profiles skip the M5/M4 boundary altogether.

## What it does

- Four new synthetic corpus entries in `tests/corpus/`, each committed
  as a `<id>.json` manifest + `<id>.body` sidecar; all four pass the
  existing corpus lint.
- A deterministic builder (`tests/harness/corpus_thresholds.py`) that
  sizes filler against the **CC-converted** shape the bridge actually
  compares (via `MessagesTranslator.translate_request`), iterates to the
  boundary, and never lets `corpus.scrub` rewrite the body.
- A regeneration script (`scripts/regenerate_corpus_thresholds.py`);
  running it twice leaves the tree untouched.
- 14 new L1 tests pinning the boundary position, scrub-inertness,
  manifest byte-identity regeneration, and structure; one new L2 test
  guarding that the README names each entry id.

## How

- **Boundary pinning is the calibration target.** Anthropic-shape and
  CC-shape messages differ by a constant ~92 chars for this layout, and
  on the `use_native_messages=True` passthrough the Anthropic shape is
  preserved verbatim. Pinning against the Anthropic shape lands 92
  chars off the boundary the bridge actually compares; pinning against
  the CC shape lands on it. The builder uses `MessagesTranslator` to
  measure what `_compact_messages` will see.
- **The over entry's filler alone crosses the threshold** so the
  post-M3-truncation body is still over — without this, M3's truncation
  of the oversized tool result drops the body back below threshold and
  M5 short-circuits instead of firing the pruning step. The asymmetry
  is covered by `test_over_budget_still_over_threshold_after_m3_truncation`.
- **M5 is not declared in the manifest.** It is profile-dependent
  (the static threshold is one budget; profile-derived `messages_budget`
  is another; on smaller-budget profiles the under entry is a TRIGGER
  case, on larger ones it's a complement). Oracle slices that consume
  these entries declare M5 at the call site that resolves the profile.
- **Human review of the 2.8 MB padded body is replaced** by a
  committed, byte-identity regeneration test (the owner of the ticket
  explicitly authorised synthesis; per `tests/corpus/README.md`
  §"Two things this format does not decide", the format imposes no
  size ceiling).
- **The over entry's tool_use carries `/tmp/…`** rather than
  `/home/<user>/` so the scrubber's `home_path` rule does not rewrite
  the committed bytes between builder output and the committed file.

## Review / verification

- `tests/harness/test_corpus_thresholds.py` — 19 L1 tests green.
- `tests/harness/test_corpus_lint.py` — 13 L2 tests green (existing
  fixture-load + lint-clean + the new README-names-entries guard).
- `tests/harness/` — 1502 tests green.
- `ruff check .` — clean.
- `mypy src/kitty` — clean.
- `lint-imports` — 5 contracts kept.
- **Two cycles with the `code-reviewer` subagent**, each producing
  blocking findings (Anthropic-vs-CC pinning; orphan tool_result defect
  surfaced during NF2 implementation) — both fully addressed; final
  re-review verdict **APPROVE**.
- **System-design-reviewer** flagged the static-constant framing on the
  first round; the chosen interpretation is **Pin the CC-converted
  shape (Recommended)**, and the README + manifest now state the
  calibration honestly.


shelpuk:
**Addressed both findings from Claude Code Review round 3** — commit `fe1a09f`.

## 1. Stale `server.py:NNNN` line-number citations (Warning)

Every citation the review named — `_safe_size`, the static-fallback branch, the short-circuit comparison, the `_get_max_context_chars` / `messages_budget` arithmetic, `_truncate_oversized_tool_results`, `_truncate_oversized_responses_outputs` — is now cited by **symbol**, matching `.system_design/SYSTEM_DESIGN.md`'s stated convention ("Symbols are named, not line-numbered"). Touched in four places: `tests/corpus/README.md`, the manifest metadata in `tests/harness/corpus_thresholds.py` (`_ENTRY_METADATA`), the builder docstrings, and `_apply_m3_truncation`'s docstring (whose `7267-7294` / `7314-7325` ranges the review correctly noted had drifted into `_validate_tool_call_pairing`'s neighbourhood).

The four committed `.json` files were regenerated via `scripts/regenerate_corpus_thresholds.py` so the manifests carry the corrected origin_note text. The byte-identity regeneration test stayed green against the new metadata.

## 2. Dead branches in `_filler_only_messages` (Suggestion)

Rewrote the helper as a shape-based filter — drop `assistant`-with-`tool_calls` (the tool_use) and drop `role=="tool"` (the tool_result) — instead of content-matching a "Reading the file now." turn that no longer exists in the over entry, and a `role=="user"` list-content check that never fires post-CC-conversion. Output on the over entry is byte-identical to before; the removed branches never fired.

## Verification

48 corpus-side tests green, 1 842 harness + layer tests green, ruff clean. CI dispatched on `fe1a09f`.

PR #160, opened 2026-09-14
test: README to code table guards (KBR-76, T-G1)

shelpuk:
## Context — for the Product Owner

Kitty Bridge's README is the product's interface specification: IDE and custom-script users configure against the bridge-mode endpoint table, debug routing via the `X-Kitty-*` attribution-header table, enable logs with the logging-flag table, and set proxy/summary paths via `KITTY_*` environment variables. Four user-facing surfaces, four corresponding code artefacts — and until this PR, no test held any of them in agreement. A drifted README reads as a product defect to the user who follows it (that is exactly what happened in KBR-9, where the README documents a Gemini endpoint the bridge does not serve).

This PR delivers ticket **KBR-76 (plan task T-G1)**: four L2 contract guards in one file, `tests/test_readme_table_guards.py`, keeping the four README tables in sync with the code. No product code changed — this is test infrastructure plus one registry row.

## What landed

**`tests/test_readme_table_guards.py`** — 21 tests, `pytestmark = pytest.mark.l2`, four guard classes:

1. **Endpoint table** — README "Available endpoints" ↔ `BridgeServer._register_routes` (bridge-mode branch + the unconditional `/healthz`//`stats` registrations; launch mode is out of scope). AST-scanned by function name, so a refactor that relocates route registration fails loudly rather than silently yielding an empty set.
2. **Attribution headers** — README `X-Kitty-*` table ↔ `BridgeServer._attribution_headers`, plus a negative arm: no file under `src/kitty/` other than `bridge/server.py` may contain an `X-Kitty-*` literal. The leak scan runs over raw source text (case-insensitive regex), so a mention inside a comment — the shape a lazy hand-off takes — is caught; a falsification case pins exactly that.
3. **Env-var register** — every `KITTY_*` env var the README names (`KITTY_EGRESS_PROXY`, `KITTY_SESSION_SUMMARY`) must exist as an **exact-match** string constant under `src/kitty/`. Forward direction only; the README is silent on internal keys (`KITTY_TMUX_WRAPPED`, `KITTY_THEME`, …) by design. The exact-match semantics are pinned by their own falsification case (a substring-only match does not count).
4. **Logging-flag table** — README "Logging" flag table ↔ the four flags in `kitty.cli.main._build_parser` (public `parse_known_args` API, no argparse internals) plus the two default paths (`BridgeServer._DEBUG_LOG_PATH` module constant; per-instance `_usage_log_path` via a stubbed `BridgeServer`, mirroring `tests/test_cli_log_file.py`). An unknown-flag negative control proves the accept-tests are not vacuous.

**`tests/exemptions.py`** — one row: `t-g1-endpoint-table` (`issue="KBR-9"`), wrapping the endpoint-table agreement assertion in `ratchet(...)`. This is plan §16's exemption path: the guard covers four tables, broader than the single KBR-9 defect, so the known drift lands exempted and **KBR-9's README correction stays in its own ticket**. When that correction lands, the exempt assertion begins passing, `UnexpectedExemptionPass` fails the job, and the row must be deleted — the T-W7 mechanism working as designed.

**Self-guards + falsifications.** Every guard asserts its own scan found known positives (the six registered routes, the three minted header names, the two documented env vars, the four parser flags), so none can rot into a vacuous green. Every pure checker has a falsification case per implementation-plan §1.4: a deliberate defect fed under controlled inputs is caught.

## Evidence the endpoint guard fails at the base revision (AC-R1.2)

Ran `check_endpoint_agreement(readme, code)` standalone at the base revision (before the exemption row existed):

```
README documents `POST /v1/gemini/generateContent` but no bridge-mode route exists for it
bridge mode registers `GET /v1/models` but the README endpoint table omits it
bridge mode registers `POST /v1beta/models/{model:.*}:generateContent` but the README endpoint table omits it
bridge mode registers `POST /v1beta/models/{model:.*}:streamGenerateContent` but the README endpoint table omits it
```

That is exactly KBR-9's planned correction scope ("Correct the table: both `/v1beta/models/{model}:generateContent` and `:streamGenerateContent`, and add `GET /v1/models`"). With the ratchet row absent, the test fails with precisely this diff; with the row present it passes; when KBR-9 lands, the row itself fails the job.

## Verification

- New tests: 21, all green (`pytest -m l2 -q tests/test_readme_table_guards.py`).
- T-W7 mechanism untouched: `tests/test_assertion_exemptions.py` 33/33 green.
- Repo gates: `ruff` clean on both touched files, `mypy src/kitty` clean (94 files), `lint-imports` 5/5 contracts kept.
- No product code changed; no other test file touched.

## Design decisions worth reviewing

1. **Exemption path over atomic fix** — the ticket records the choice explicitly; §16's defect table lists KBR-9 as "atomic fix + test now", which is the route for KBR-9's own ticket, while T-G1 (the broader guard) takes the exemption path. Recorded in the requirements doc's "Reconciliation with §16's defect table" paragraph.
2. **Raw-text scan (not AST) for the X-Kitty leak arm** — an AST walk would miss a comment-side leak; the docstring-plus-comment falsification pins the mechanism choice.
3. **Exact-match (not substring) for the env-var register** — a docstring mentioning `KITTY_SESSION_SUMMARY` must not count as a reader; the falsification pins it.
4. **Forward-only env-var direction** — the README legitimately omits internal keys; an exact-equality guard would force documenting plumbing or renaming it, neither of which is right. (The §6.2.3 env-var row's `ANTHROPIC_*` register is a different contract, already pinned code↔code.)

## Tickets

- Closes **KBR-76** (T-G1) when merged.
- Relates to **KBR-9** (the endpoint-table drift; the exemption row names it and auto-fails when it closes).


shelpuk:
Thanks for the review — both findings addressed in commit `645db23`:

* **Docstring accuracy** (`tests/test_readme_table_guards.py:28`): corrected the named constant from `_DEBUG_LOG_DIR` to the actual contract — `_DEBUG_LOG_PATH` for the debug path (imported) plus the per-instance `_usage_log_path` default (which constructs `_DEBUG_LOG_DIR / "usage.log"` internally). The constant `_DEBUG_LOG_DIR` does live in `bridge/server.py`, so the docstring now names both for accuracy.
* **Dead constants** (`tests/test_readme_table_guards.py:372`): dropped `_ATTRIBUTION_HEADER_NAMES`, `_DEFAULT_USAGE_LOG_FRAGMENT`, `_DEFAULT_DEBUG_LOG_FRAGMENT` — leftovers from an earlier design sketch that ended up unreferenced.

No behaviour change: 21/21 new guards + 33/33 T-W7 mechanism tests still green; ruff clean. The Windows leg is still pending from the previous CI run; a fresh run will exercise the cleanup commit once GitHub re-queues it.

PR #161, opened 2026-09-15
T-J1 (KBR-107): pytest-bdd wiring and the L4 acceptance layer

shelpuk:
## Status (T-J1 / KBR-107)

**Implementation: shipped** — round 2 review fixes (commit `24b64d5`) addressed the one warning and three suggestions the Claude Code Review posted on this PR.
**CI: all checks green** — the 6 test legs across Linux/Windows/macOS, CodeQL, Claude Code Review, and `ci-required` all SUCCESS.
**Status: MERGEABLE / CLEAN** — ready for human review and merge.
**Not merged by author** — per project convention; awaiting human sign-off.
**Jira:** KBR-107 has a full implementation comment posted.


shelpuk:
## Round 2 — addressing this review

**The warning (verified real, then fixed).** I checked `BridgeFixture.stop()` against `__aexit__` before touching anything: the reviewer is right — `stop()` (bridge.py:760) releases ports but never runs `assert_teardown_clean()`; only `__aexit__` does, on the clean path. `bridge_session` called `stop()` directly, so the smoke never exercised the contract's fourth assertion, and my twin helper's docstring claimed otherwise. Fixed in `24b64d5`:

- `tests/acceptance/conftest.py` now mirrors `__aexit__` exactly — `try: yield / except: stop, close, re-raise / else: stop, assert_teardown_clean, close` — so the declared-format check runs on the clean path and is skipped when the test body itself failed, same as the L3 reference. The docstring lie is corrected in `tests/harness/bridge.py`.
- This is precisely the §1.4 trap the review named, and it was caught because the falsification suite for the *original* helper documents that teardown is assertion 4 — the wiring test inherited the blind spot from the helper it called.

**Suggestion (a) — applied.** `_honest_run()`'s `try`/`finally` only wrapped the assertion; a raise from the drive leaked a started fixture. Replaced with an `async with _started()` context manager so the entire body — start, drive, assert — is inside one teardown guarantee.

**Suggestion (b) — applied.** The single `finally` in `bridge_session` ran `stop()` then `close()`; a raise from `stop()` skipped `close()`. The `except`/`else`/`finally` shape closes the loop on every exit path.

**Suggestion (c) — applied.** The docstring's "order is load-bearing" claim was wrong: both `scenarios()` and the step decorators run at module import, before collection, and the first working version of this file had the decorators first. The docstring now states the actual load-bearing property — steps must live in a module pytest collects (`test_*.py`), because pytest-bdd injects step fixtures into the calling module's namespace. The first version of this PR shipped steps in `steps.py`, which pytest never collected, and every step resolved as "not found" — that is the failure the (overstated) comment was trying to warn the next author about.

**Verification after the fixes:** `pytest -m acceptance -q` 1 passed; `tests/harness/test_bridge_falsification.py` 10 passed; `tests/test_layer_markers.py` + `tests/test_layer_selection.py` 70 passed; `ruff check .`, `lint-imports`, `mypy src/kitty` all clean.

shelpuk:
## Round 3 — addressing this review

**Warning (verified).** Read `BridgeFixture.start()` against my conftest again — yes, `start()`'s two guards (`bridge.py:707-732`) raise `BaseException`, and my `loop.run_until_complete(bridge.start())` was outside the `try`, so a profile-validation or bind failure stranded the loop with no `close()`. Fixed in `0026e7d`: everything now runs inside one `try` whose `finally` closes the loop on every exit. The inner `try`/`except`/`else` still mirrors `__aexit__` (stop on raised, stop-then-assert on clean).

**Suggestion (docstring).** "The L3 harness surface shipped in `tests/harness/` and `tests/exemptions.py`" implied `exemptions.py` was already imported here; it isn't (the TR-1c exemption row lands with T-J2). Reworded to name `harness.bridge` as the surface this module actually binds to, and recorded the exemptions hook as a T-J2 follow-on, so a reader does not grep for an import that isn't there.

**Suggestion (§1.4 — the substantive one).** The conftest's `transport_instance.assert_teardown_clean()` had no falsification. Extracted the clean-path teardown into `tests/acceptance/teardown.py::teardown_clean_path` (plain function, not fixture-internal) so both the conftest and a falsification test call the same helper; the falsification passes a transport whose `assert_teardown_clean` raises `MisdeclaredFormatError` (mirroring `test_bridge_falsification.py`'s `_MisdeclaredTransport`) and wraps the call in `pytest.raises`. The helper had to be a plain function: fixture teardown errors surface as test errors, not as exceptions a test can wrap — a helper only the fixture calls cannot be falsified.

A regression that drops or reorders the assertion now goes red; the gate has a detector instead of a decoration.

**Verification after the fixes:**
- `pytest -m acceptance -q` → 2 passed (the smoke + the falsification).
- `--layer-report` confirms the new file carries `["acceptance"]` via the path default — same layer, no hand-marker needed.
- `tests/harness/test_bridge_falsification.py` → 10 passed (unchanged).
- `tests/test_layer_markers.py` + `tests/test_layer_selection.py` → 70 passed.
- `ruff check .`, `lint-imports`, `mypy src/kitty` all clean.

shelpuk:
## Round 4 — addressing this review

All four suggestions applied in `d3a2361`.

**§1.4 chain (the substantive one).** You're right that the chain had a hole, and the honest fix is not the one the docstring claimed. I looked at closing it end-to-end and the options were: (a) `pytester`'s nested-run pattern — the only way to `pytest.raises` a *fixture's* teardown error, since fixture teardown failures surface as test errors the test body cannot catch; or (b) a hook/flag-based transport — ugly and fragile. Taking (a) meant either copying the conftest into a pytester tmpdir (which tests a *copy*: the real conftest could regress while the nested copy still passes) or `pytester.copy_example` (which pulls repo-relative fixtures into a tmp tree — more machinery than this seam warrants at T-J1, when the layer does not gate until T-K9). So the fix I took is the honest smaller one: **removed the dead `request.param` plumbing** (no test parametrised it) and **reworded the conftest docstring** to state the actual coverage boundary — the helper is falsified directly; a regression that deletes the conftest's one-line `else` call (rather than edits the helper) is *not* runtime-caught here, and pytester is named as the route if that coverage becomes important at T-K9. The boundary is now on the record in the code, rather than a false claim in a docstring.

**`transport_instance` parameter.** Dropped; `teardown_clean_path` reads `bridge.transport`. Both call sites (conftest `else`, falsification) updated.

**`_MisdeclaredTransport` message.** You're right — the scenario never drives a request, so "body was gemini" asserted a body that did not exist. The message now says what the transport *is* ("the deliberate mis-declared-format defect") and the test asserts on that string, so the message and the assertion cannot drift apart silently.

**`__all__`.** `assert_fixture_reached_its_recorder` added alongside its sibling.

**Verification:** `pytest -m acceptance -q` 2 passed; `tests/harness/test_bridge_falsification.py` 10 passed; layer guards 70 passed; `ruff check .`, `lint-imports`, `mypy src/kitty` all clean.

shelpuk:
## Round 5 — addressing this review

**Suggestion (bridge.py:1016).** Right — and it broke the docstring's own promise ("the same failure vocabulary as the original"). Fixed in `4f89ebd`: the helper now takes a `body` keyword and its third-assertion message ends with `body[:200]!r` exactly as the original's does. The smoke step passes the reply body it already captured through to the helper (no second read), and the four twin falsifications pass bodies that match what the test claims happened ("ok" for the honest path, "server exploded" for the wrong-status one) so a red run reads correctly.

Everything else in this review — the loop-closure audit, the §1.4 chain boundary as a documented trade-off, the two settled docstring claims — matched my reading of the code. No other action taken.

**Verification:** `pytest -m acceptance -q` 2 passed; `tests/harness/test_bridge_falsification.py` 10 passed; layer guards 70 passed; `ruff check .`, `lint-imports`, `mypy src/kitty` all clean.

shelpuk:
## Review closed — five rounds, 9 findings applied, no new findings

The Claude Code Review returned **no findings** on the round-5 head (`4f89ebd`). Each prior round's fix was verified in the diff (round-2 loop leak, round-3 helper extraction / §1.4 falsification / dead-`request.param` removal / `__all__`, round-4 third-assertion message). The §1.4 chain partial is the trade-off the prior rounds explicitly accepted; it is documented on the record in both module docstrings with `pytester`'s nested-run pattern named as the route to end-to-end conftest coverage when it becomes important (T-K9).

The 6 test legs across Linux 3.10–3.13 / Windows / macOS are still running; I'll wait for them to finish before reporting final state to you.

Issue #22, opened 2026-08-21
Concurrent sessions: launching a second kitty claude session reroutes every already-running session via Claude Code settings hot-reload


shelpuk:
## Summary

Claude Code **watches its settings files and hot-reloads them into running sessions** — the `env` block is re-applied to the process environment of a running session every time the file changes. Because every `kitty claude` session patches the same user-global `~/.claude/settings.json`, **each new session launch rewrites every other running session's `ANTHROPIC_BASE_URL` / `ANTHROPIC_API_KEY` / `ANTHROPIC_AUTH_TOKEN` to the new session's bridge and key**.

Sessions running different profiles (common when worktrees use different providers/models) get silently re-routed through the newest session's bridge → wrong-provider responses, auth/entitlement errors, or model-not-found failures **mid-session**. This matches the user-visible report: "multiple agents running; one session started erroring when another session started".

This is the architectural defect of the shared-file design: the single `ANTHROPIC_BASE_URL` key can only hold one session's value, and any write hot-reloads into all running sessions.

## Key use case

Multiple Claude Code agents running on the same PC with kitty — the configuration (`~/.claude/settings.json`, `~/.config/kitty/*`) is shared — but working in different worktrees.

## Environment

- kitty-bridge 1.4.0 (`main` @ cc16213), Claude Code adapter
- OS-independent; egress proxy configuration is **not** required to reproduce

## Verified Claude Code behavior (official docs, checked 2026-08-21)

> "Claude Code watches your settings files and reloads them when they change, so edits to most keys apply to the running session without a restart." — [settings.md](https://code.claude.com/docs/en/settings.md)

> "Claude Code writes each `env` entry into the process environment at startup and again when the file changes, replacing the value inherited from the shell." — [env-vars.md](https://code.claude.com/docs/en/env-vars.md)

Relevant nuances:

- Additions/changes propagate to running sessions; **removals do not unset** variables in a running session (next launch only).
- `ANTHROPIC_MODEL` (and the `model` key) apply only to the session launched with them — cannot be switched by editing settings mid-session.
- Whether `ANTHROPIC_BASE_URL`/`ANTHROPIC_API_KEY` are re-read per request is not explicitly documented, but `apiKeyHelper` is explicitly hot-reloaded and env values are rewritten into the process environment — treat mid-session rerouting as likely (observed failures are consistent with it).

## Reproduction

1. Start session A in worktree A on profile P1 (provider X)
[...]

PR #162, opened 2026-09-15
KBR-93: settings lifecycle (T-I1) — six L3 subsystem tests

shelpuk:
# KBR-93 / T-I1 — Settings lifecycle

## Context

`kitty claude` does not rewrite the user's global `~/.claude/settings.json`
since issue #22: it writes a per-session settings file (under the OS temp
dir) and points Claude Code at it via `--settings`. The user's global
file is read once — for a stale-value warning — and never written. Three
ways a session can end must keep that guarantee: the child exits cleanly
(the `finally` block restores), the user signals `SIGINT`/`SIGTERM` (the
same `finally` block runs after the forwarded signal kills the child),
and a hard kill leaves a recoverable orphan that `kitty cleanup` repairs
without ever touching state it does not own.

This PR delivers the L3 subsystem coverage for that guarantee — real
child processes, a fully redirected filesystem, the real
`kitty cleanup` command — and corrects two stale claims in the design
and code that the new tests surfaced:

- The SIGTERM scenario in `.system_design/TEST_SUITE.md` §6.3.2 said
  "atexit path restores." It does not: `launch_async` installs
  `signal.signal(SIGTERM, _forward_signal)` while the child runs, so
  SIGTERM is forwarded; the `finally` block is the path. The atexit
  docstring at `src/kitty/cli/launcher.py:40-55` carried the same stale
  claim. Both updated.
- The `_kitty_values_present` heuristic in `launchers/claude.py:124-148`
  fires on a loopback `ANTHROPIC_BASE_URL` **or** the kitty auth token —
  by deliberate design, documented at the function. The negative control
  test for that heuristic was originally written with a loopback URL,
  which would have exercised the wrong branch. Corrected: the negative
  control uses a non-loopback URL, and a stdout branch-pin
  (`"Removed stale backup"`) catches any future broadening.

## What lands

- `tests/cli/test_cli_settings_lifecycle.py` — six L3 tests with a
  hermetic sandbox harness. All six pass; two targeted product
  mutations (`os.unlink(original)` → `pass`; `_kitty_values_present`
  returning `True` unconditionally) are caught with legible messages.
- `src/kitty/cli/launcher.py` — docstring fix only (corrects the
  SIGTERM/atexit claim, names the pre-handler window as a known
  limitation).
- `.system_design/TEST_SUITE.md` — §6.3.2 SIGTERM row now describes
  forwarding + `finally`; heuristic row names the loopback-or-token
  semantics and the non-loopback negative control.

## Why this matters to the user

After this PR lands, every way a `kitty claude` session ends has an
L3 subsystem test proving the user's global settings file is untouched
and the per-session state is cleaned up (or, for hard-kill, cleanly
orphaned for `kitty cleanup` to handle). The user-facing contract of
issue #22 — sessions never mutate the user's global config — is now
tested end-to-end.

## L3 activation

L3 has no CI job today (plan task T-K6). The new tests run locally on
`pytest` and are listed in `tests/layers.py::PENDING_ACTIVATION_LAYERS`
as pending the Subsystem job. When T-K6 lands, these tests gate.

## Out of scope (recorded)

- `cli/cleanup_cmd.py:201-244` is dead code (duplicate heuristic block
  after an unconditional `return 0` at line 199). Noted in the Jira
  ticket; not touched per surgical-change discipline.
- `_atomic_write_text` opens with default `newline=None`, so a LF backup
  round-trips to CRLF on Windows. The byte-exact restore test (AC-4)
  is `@skipif` on Windows with a documented reason; a Windows-only
  byte-equivalent comparison or a fix to `_atomic_write_text` would
  unblock that leg.


shelpuk:
## Review fixes (commit `e64d261`)

A pre-merge review pass surfaced five findings; all applied:

| # | Finding | Fix |
|---|---|---|
| W1 | Stale line-number citations in three living documents (the doc-truth edit shipped with `line 249/308/261` references that rotted as soon as the edit landed). | Replaced with symbol citations — `_register_atexit_cleanup` call site, `signal.signal(SIGTERM, _forward_signal)` install, `build_child_env` call site. In `launcher.py:51-54`, `TEST_SUITE.md:2197`, and three test comments. |
| W2 | AC-5's atexit discriminator was vacuous: the probe child's stderr was empty because `BridgeServer` installs `_crash_excepthook` (server.py:124-136) which routes the traceback through the `kitty.bridge` logger and calls `sys.exit(1)`. A refactor moving `build_child_env` inside the `try` block would have stayed green. | The driver now calls `logging.basicConfig(filename=<sandbox>, force=True)` before any kitty import. The atexit's lone `atexit cleanup: restored` INFO line is now observable. AC-5 asserts that marker is in the captured log. **Empirically verified**: a mutant that moves `build_child_env`'s call inside the try block is now killed — the except branch emits `cleanup_launch: removed session settings` (from the finally) and no atexit marker. |
| W3 | `_Sandbox` disjointness guard covered home and cache but not the config tree. A broken `XDG_CONFIG_HOME` redirect would silently seed the developer's real `ProfileStore` + `FileBackend`. | Added `config_dir` probe and guard; one assertion, in spirit with the home/cache guards. |
| W4 | The Windows `.cmd` stub ignored its `mode` argument and always wrote `exit /b 0`, making AC-3 (the SIGKILL scenario) red-on-Windows. | The Windows stub now branches on `mode`: `ping -n 999 127.0.0.1 >nul` for `sleep`, `exit /b 0` for `exit0`. `proc.kill()` lands as `TerminateProcess` on Windows regardless. |
| W5 | `PYTHONPATH` leaked into both child kinds. | Already applied in commit `f622587`: `env.pop("PYTHONPATH", None)` in `_sandbox_env`. |

All six tests still pass; ruff clean; mypy on `src/kitty` clean; lint-imports 5/5 kept.

shelpuk:
## Disposition of the review findings

All seven findings (2 warnings + 5 suggestions) from the Claude Code review at commit `e64d261` triaged below.

### Warnings

**W-1 · `README.md:635` — README cleanup section still says SIGTERM is the atexit path.** **Fixed** in follow-up PR [#166](https://github.com/Shelpuk-AI-Technology-Consulting/kitty-bridge/pull/166). I applied the factual correction but trimmed the reviewer's verbatim suggestion: the internal symbol names (`_register_atexit_cleanup`, `signal.signal(SIGTERM, _forward_signal)`) and the pre-handler window caveat belong in the docstring and the design doc, where engineers look; the README stays user-facing. The fact-pattern (three named layers, manual `kitty cleanup` for the unrecoverable cases) is identical.

**W-2 · `src/kitty/cli/cleanup_cmd.py:244` — dead duplicate cleanup branch below an unconditional `return 0`.** **Acknowledged, not fixed in this PR.** The PR description already records this as out-of-scope per surgical-change discipline: the deliverable is the L3 settings-lifecycle coverage, not a product cleanup, and touching `src/kitty/cli/cleanup_cmd.py` for dead-code removal would have bloated the diff and put the change on a review path where it does not belong. The reviewer's own body notes "I am raising it as a warning, not a critical." The finding is recorded in the PR body, in [KBR-93's Jira comment](https://shelpuk.atlassian.net/browse/KBR-93), and in the "Out of scope" section of [KBR-260](https://shelpuk.atlassian.net/browse/KBR-260) (the follow-up HIGH ticket) so it surfaces when KBR-260 is worked.

### Suggestions

**S-1 · `test_cli_settings_lifecycle.py:829` — AC-5's stderr discriminator is too easy to silence on a refactor.** **Ack; the test already names the atexit marker as the load-bearing oracle.** The code above the assertion calls it the **primary discriminator** and the stderr check the **secondary discriminator**, explicitly documented ("the genuine atexit path's traceback goes through the logger … and not stderr, so stderr stays clean of this string"). A future change routing the traceback to stderr would indeed degrade the secondary assertion — but the primary (`atexit cleanup: restored` in the captured log file) survives that change, so the test stays load-bearing. No change; the naming already communicates the hierarchy.

**S-2 · `test_cli_settings_lifecycle.py:148` — `_RecordingClaudeAdapter` `__getattr__` delegation can hide adapter regressions.** **Ack; the trade-off is deliberate.** The wrapper is a *test-only* shim that records the session-file path for the parent's readiness barrier. The contract it proves is "the launch path runs end-to-end against the real adapter" — and any adapter regression the launch path touches surfaces as an `AttributeError` in the child, which fails the test loudly. The silent-regression case the reviewer sketches (a lazily-populated attribute returning `None`) is real but is exactly the kind of adapter change that L1 coverage in `tests/test_launcher_claude.py` would catch; the L3 file is not the layer where adapter surface is proven. No change.

**S-3 · `test_cli_settings_lifecycle.py:367` — `_Sandbox` disjointness guards are path-equality, not real containment.** *(low confidence from the reviewer)* **Declining.** The path-equality guard is the containment contract the sandbox actually needs: the redirected env makes platformdirs and `Path.home()` resolve *literally* into the sandbox, so a `~/.config/kitty` symlink into the sandbox would make `Path.home()` resolve *there* and the equality guard would still hold. The reviewer's scenario requires a deliberate symlink layout to defeat, which the sandbox itself creates — so the guard as written is not weaker than the redirect it is checking. No change.

**S-4 · `test_cli_settings_lifecycle.py:596` — the `dup2` ordering fix is correct but undocumented.** *(low confidence)* **Declining.** The comment already names the failure mode and the resolution ("passing the same file object as both stdout and stderr hits a Popen dup2 ordering quirk where the second dup wins and the first stream ends up wherever fd 1 started, which loses half the child's output"). The Python versions it holds on are the versions the suite runs on (3.10–3.13), which the Fast gate enforces; re-stating that in a single comment adds maintenance surface without new information. No change.

**S-5 · `test_cli_settings_lifecycle.py:65` — `_CHILD_TIMEOUT_SECONDS` ceiling is 120s, much higher than the actual work.** *(low confidence)* **Declining.** The 120s is a backstop, not a budget — the comment above it says exactly that, and the run history shows every scenario completes well under a second. The trade the reviewer sketches (masking a 30s hang as "real-but-slow") does not survive reading the assertion: the readiness barrier at `_READY_TIMEOUT_SECONDS = 30.0` already fails the test if the child has not reached `proc.wait()` in 30s, so a hang inside 120s is caught long before the subprocess ceiling fires. Lowering the ceiling would tighten the wrong knob. No change.

### Also noted (not actionable)

Four of the seven reviews on this PR have empty bodies — bot artifacts of the same review's sub-outputs, no content to address.

The orphaned-findings comment (`claude-code-review-orphaned`) repeats W-1 and W-2 under `rules/documentation.md` / `rules/launchers.md`; both are triaged above.

[KBR-260]: https://shelpuk.atlassian.net/browse/KBR-260?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #142, opened 2026-09-14
fix(streaming): re-dispatch across transport classes on plain→custom failover (KBR-249)

shelpuk:
## Context for Product Owner

A streaming `/v1/messages` request whose initial draw lands on a plain-HTTP backend was failing over to a custom-transport backend (bedrock / openai_subscription) and then driving the new provider over plain HTTP — because the bridge picks which protocol to speak **once, at the start of the request**, from the initial provider's class. The custom-transport backend answered `200` with a contentless body, and the client got a silent empty turn. The KBR-235 emptiness ladder ended the turn in a `D4` `502`. This was the path that landed all six CI legs red on PR #131 (see `tests/bridge/test_balancing_server.py::TestBalancingAllCustomTransport::test_streaming_skips_backends_without_stream_request`, and the two exemption rows that ship there as `kbr-249-failover-plain-post-{status,sse}`).

The fix dispatches across the two transport branches inside `_stream_messages`. The shape is one `while True:` loop around the existing custom-transport branch and the existing plain-POST branch. The custom-transport branch's symmetric `custom → plain` fall-through already exists (it becomes a `continue` of the new loop). The reverse direction — plain → custom — is what lands here: each of the seven pre-emission plain-POST failover sites now carries a cross-class check that breaks the inner retry and sets a flag the post-try routing reads and continues with. A per-request hop cap of `(2 * n_backends) + 1` bounds pathological cooldown ping-pong and surfaces a JSON 502 with a real message when hit.

## Diff scope

- `src/kitty/bridge/server.py` — `_stream_messages` only:
  - Dispatch loop wrapper (10 lines);
  - One `continue` on the existing custom → plain fall-through;
  - Seven cross-class detection blocks at the seven pre-emission plain-POST failover sites (`src/kitty/bridge/server.py` original line numbers 4330 / 4433 / 4581 / 4785 / 4880 / 4905 / 5067);
  - Per-iteration flags (`_cross_mode_to_custom`, `_cross_cap_hit`) and post-try routing (`if _cross_mode_to_custom: continue; if _cross_cap_hit: return _make_error_response(... cap ...); break`);
  - Eleven `E501` line-length fixes that the +4 re-indent pushed over the 120-char limit (purely cosmetic / string splits).
- `tests/bridge/test_balancing_server.py`:
  - Removed the two `with ratchet(...)` blocks around the previously-exempted `assert resp.status == 200` and `assert resp.content_type == "text/event-stream"` lines;
[...]

PR #164, opened 2026-09-15
KBR-254: cross-class re-dispatch in the three sibling stream handlers

shelpuk:
## Summary

KBR-254 extends the KBR-249 cross-class dispatch-loop fix from `_stream_messages` to the three sibling stream handlers — `_stream_responses`, `_stream_gemini`, and `_stream_chat_completions`.

Before this change, each sibling's plain-POST retry loop selected a transport class once at handler entry. When a failover (`_select_backend()`) picked a `use_custom_transport` backend mid-stream, the plain-POST branch drove it with `session.post(...)`, which the provider cannot serve — the request silently degraded to an empty response. On a balancing pool that mixes plain and custom-transport backends, `_require_streaming` makes such crossings inevitable.

## What changed

Each of the three handlers now carries the same dispatch-loop shape KBR-249 put on `_stream_messages`:

- A `while True:` dispatch loop re-evaluates the selected provider's transport class per iteration, so a `plain → custom` crossing re-enters the custom-transport branch with the failover's own selection (no second `_select_backend()` call — the two-draw invariant pinned by `tests/bridge/test_balancing_server.py` is preserved).
- A per-request hop cap of `(2 * n_backends) + 1` bounds pathological cooldown-expiry ping-pong and surfaces a defined terminal event instead of looping.
- Nine plain-POST failover sites carry the transport-class crossing guard (three per handler: non-2xx, in-stream error on the no-events path, and the empty-response ladder). The cloudflare sites abort without failover on all three siblings, so they carry no crossing and are not guarded.

Six new L1 tests in `tests/bridge/test_balancing_server.py::TestBalancingAllCustomTransport` prove each handler's fix: one content-asserting positive test per route (failover onto a custom backend re-dispatches and the mocked payload reaches the client — not the KBR-249 vacuous-oracle shape), and one cap-hit test per route. The initial backend draw is pinned per `mem:backend-selection-is-random-among-primaries`; the cap-hit tests use the KBR-249 cap-test mechanism (alternating draw, `_mark_backend_unhealthy` no-op, `_any_healthy_backend(require_streaming=True) -> False`).

## Per-route cap-hit terminal shape (the deliberate asymmetry)

`/v1/messages` defers `sr.prepare()` until the first real write, so KBR-249 answers a cap-hit there with a bare JSON `502`. The three siblings prepare their SSE response eagerly at handler entry, so the status line is already committed — the cap-hit has to surface in-stream, in each route's own D4-family shape:

| Route | Terminal event | Discriminator |
|---|---|---|
| `/v1/responses` | `event: error\ndata: {"type": "error", "sequence_number": N, "code": "cross_class_exhaustion", "message": "...", "reason": "cross_class_exhaustion"}` then the existing `synthesize_completed_events(status="incomplete")` + `write_eof` | `code` (+ `reason`) |
| `/v1beta/...:streamGenerateContent` | `data: {"error": {"code": 502, "message": "...", "reason": "cross_class_exhaustion"}}\n\n` then `write_eof` | `reason` |
| `/v1/chat/completions` | `data: {"error": {"message": "...", "type": "cross_class_exhaustion"}}\n\n` + `data: [DONE]\n\n` then `write_eof` | `type` |

The message body is identical to the Messages route's KBR-249 wording: *"Upstream backends exhausted (the bridge could not land on a usable backend on this request)"*. The discriminator follows each route's established D4 family (`code` on Responses since KBR-247, `reason` on Gemini since KBR-250, `type` on Chat Completions since the empty-response arm).

The asymmetry is recorded in `.system_design/SYSTEM_DESIGN.md` §5.4.

## Test evidence

- 6 new L1 tests pass (Responses / Gemini / Chat Completions, positive + cap-hit each).
- Full `tests/bridge/test_balancing_server.py` suite (59 tests) passes — includes the existing KBR-249 tests and the KBR-247 exhaustion-arm guard.
- `ruff check .`, `mypy src/kitty`, `lint-imports` all green.
- Full `pytest -m "l1 or l2"` fast gate — see CI.

## Traceability

- Ticket: [KBR-254](https://shelpuk.atlassian.net/browse/KBR-254)
- Design: `.system_design/SYSTEM_DESIGN.md` §5.3 S8/S9 and §5.4
- Requirements: `.requirements/20260915T113835Z_kbr_254_sibling_cross_class_dispatch/REQUIREMENTS.md`
- Parent fix: KBR-249 (PR #142)

## Context for the product owner

A balancing pool can mix two kinds of upstream backends: ones the bridge talks to with plain HTTP (Chat Completions-style), and ones with vendor-specific custom transports (e.g. Bedrock over boto3). When the pool is mixed and a request's first backend fails, the bridge can fail over onto the *other kind* of backend — and before this fix, the code that handles the request had already committed to speaking the first backend's protocol, so the failover silently produced an empty reply instead of a real answer. KBR-249 fixed this on the `/v1/messages` route. This change extends the same fix to the other three routes the bridge serves (`/v1/responses`, `/v1/chat/completions`, and Gemini), so a mixed pool fails over correctly on every route. It also gives each route an honest, client-visible error when the pool has bounced the request between backend kinds so many times that the bridge gives up, instead of the empty answer users would have seen.

[KBR-254]: https://shelpuk.atlassian.net/browse/KBR-254?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


shelpuk:
Round-2 push addressing the Claude Code Review warnings:

**§5.3 S8 vs §5.4 Chat Completions on `reason`** (design-drift + test-coverage): resolved by keeping Chat Completions's discriminator as `type` alone (which is what the implementation and §5.4 already say) and rewording §5.3 S8 to say so explicitly — each route's cap-hit carries *its own* D4 discriminator (`code` on Responses, `reason` on Gemini, `type` on Chat Completions), and Chat Completions does **not** carry `reason`. Per the reviewer's own closing note, no test change is needed for Chat Completions under this reading.

**Broken indentation on the three `_mark_backend_unhealthy(...)` splits** (simplification): arg line moved from column 32 to column 52 and the closing paren from column 28 to column 48, so the closing paren now sits at the opening paren's column and the arg line is one indent deeper — consistent with the surrounding 48-space block. All three sites (Responses in-stream, Gemini in-stream, Chat Completions in-stream) got the same fix.

**Cap-hit `logger.warning(...)` over-indent** (style): closing paren moved from column 52 back to column 44 at all nine guard sites so it aligns with the `logger.warning(` opener. (These are the *new* KBR-254 guard blocks; the pre-existing `_stream_messages` blocks the suggestion listed were not touched.)

shelpuk:
Round-3 push addressing the round-2 Claude Code Review warnings:

**§5.4 stale "defect exists" bullet** (design-drift): merged the pre-existing bullet — which said the sibling defect still exists and "A sibling ticket per handler is owed" — into the KBR-254 bullet. The merged bullet keeps the pre-KBR-254 site coordinates per handler (now historical), and tells the next reader to grep `KBR-254: cross-class failover` to locate the actual guarded sites in current source. Also dropped the duplicate KBR-254 bullet this PR added in round 1; its content is now in the merged bullet.

**README troubleshooting entry** (doc-drift): the entry claimed the cap-hit is "Balanced profiles only, streaming `/v1/messages` only" and described it as a bare 502. Rewrote it to name all four streaming routes and the per-route terminal event shape (Responses `code` + `reason`, Gemini `reason`, Chat Completions `type`, plus the standard message wording) — and explained the intentional 200-on-the-wire asymmetry on the three sibling routes, which prepare their SSE response eagerly and cannot rewrite the status after the fact.

**Residual substituted-arg over-indent** (style): the round-2 commit fixed the closing paren but left `_crossings, _max_crossings,` one indent deeper than the format-string args. All nine KBR-254 cap-hit guard sites now have the substituted-arg line at the same indent as the format strings.

shelpuk:
Round-4 push addressing the round-3 Claude Code Review warnings:

**Messages catch-all cap-hit block under-indented `message_id,`** (style-regression): the round-3 dedent of substituted-arg lines matched the KBR-249 cap-hit block too (its format string starts with `Re-dispatch cap reached...`) and knocked `message_id,` from 20 spaces down to 16, leaving the closing `)` at 20. This round restores both to 20 — same column as the format strings. My round-3 sed-style fix should have excluded the pre-existing KBR-249 block; that was my error.

**Responses catch-all: `response_id,` merged onto the `_max_crossings,` line** (code-style): this was a wrap-script artefact from round 1 — my `{log_id_arg}` template slot had no leading newline, so the two substituted args got concatenated with ~20 spaces of padding, and the blank line that followed sat before the closing paren. Split onto two lines at the correct indent.

**Gemini + Chat Completions catch-all: blank line before the closing `)`** (code-style): same root cause — the `{log_id_arg}` slot was passed `"\n"` for routes without a third substituted arg, leaving an empty line between the args and the paren. Removed.

**Chat Completions cap-hit test does not pin that `reason` is absent** (test-coverage): agreed — after the round-2 §5.3 S8 clarification (Chat Completions's D4 discriminator is `type` alone, no `reason`), the absence is now part of the contract a client can rely on. The `_run_cap_hit` helper grew an `absent_discriminators` parameter, and the CC test passes `absent_discriminators=("reason",)`. Mutation-verified: injecting `reason` into the CC cap-hit payload fails the test; removing it passes.

shelpuk:
Round-5 push addressing the round-4 Claude Code Review suggestion:

The `_run_cap_hit` helper grew a `required_discriminators` parameter — the symmetric counterpart to `absent_discriminators` (added in round 4). The Responses cap-hit test now passes `required_discriminators=("reason",)` to pin the second field §5.3 S8 promises (the parent KBR-241 marker alongside the route's D4 `code`).

**Mutation-verified.** Dropping `reason` from the Responses cap-hit payload (server.py:4048) makes the test fail; restoring it makes it pass. The same pattern would catch the symmetric regression on a future route — e.g., if Gemini ever drops `reason` for its D4, `required_discriminators=("reason",)` on the Gemini test catches it.

shelpuk:
Round-6 push addressing the round-5 Claude Code Review suggestion:

The `_run_cap_hit` helper's Google-style `Args:` block omitted the two keyword parameters added in rounds 4 and 5 (`absent_discriminators` and `required_discriminators`). Each is now documented with its semantics — `absent_discriminators` pins the per-route asymmetry §5.3 S8 promises (e.g. Chat Completions carries `type` alone, never `reason`); `required_discriminators` pins the additional fields the design doc names (Responses carries `reason`, the parent KBR-241 marker, in addition to its D4 `code`). No behaviour change.

PR #166, opened 2026-09-15
README: clarify SIGTERM lifecycle — forwarded to agent, finally path runs

shelpuk:
Completes the SIGTERM cleanup-lifecycle correction that KBR-93 (commits `f622587` + `e64d261`) started but which KBR-78 (`e66605b`, "feat(kbr-78): T-G9 exact-set header contract per adapter") reverted during its merge to `main`. The Claude Code review on this PR caught that I claimed the launcher.py and `TEST_SUITE.md` corrections already existed in the original PR — the corrections existed at PR #162's merge, but `e66605b` clobbered them. Reapplies the corrections with the symbol-citation rewording KBR-93 landed in `e64d261`, so the re-application does not reintroduce the stale `line 249/308/261` references the reviewer flagged in #162.

Three findings from this PR's Claude Code review, all addressed:

| # | Finding | Fix |
|---|---|---|
| W-1 | `launcher.py:43-45` docstring still claims SIGTERM triggers atexit | Restored the KBR-93 wording with symbol citations; appended the pre-handler window note. |
| W-2 | `TEST_SUITE.md:2198` SIGTERM row still claims "atexit path restores" (also clobbered the heuristic row at :2199 in the same revert) | Restored both rows to the KBR-93 wording: SIGTERM forwarded+finally+pre-handler window, heuristic loopback-or-token + non-loopback negative control. |
| S-1 | README "Three layers of cleanup" intro precedes a four-item list | Reworded to "Three automatic cleanup layers, plus manual recovery:" — matches the count and signals the fourth bullet is a manual walk-back. |


shelpuk:
Disposition of the three review findings — commit `d357b3a` on this PR:

**W-1 (launcher.py:43-45 docstring) — Fixed.** The reviewer was right: the file had reverted to the pre-KBR-93 stale wording. Restored to the KBR-93 wording with symbol citations (which I had already updated to in `e64d261`; my PR description was wrong about both files already being correct). The reviewer's suggestion text matches my original wording almost verbatim; I used the suggestion as the new content.

**W-2 (TEST_SUITE.md:2198 SIGTERM row, and the heuristic row at :2199) — Fixed.** The same revert also took out the heuristic row I had corrected in `f622587` (which named the loopback-or-token semantics and the non-loopback negative control). Both rows now match KBR-93's corrected wording. The reviewer only flagged the SIGTERM row; the heuristic row was collateral from the same revert, so I re-applied both — documenting here for transparency in case the reviewer prefers a separate commit. Used the reviewer's suggestion text for the SIGTERM row.

**S-1 (Three layers of cleanup vs four-item list) — Fixed.** Adopted the reviewer's suggested wording verbatim ("Three automatic cleanup layers, plus manual recovery:"). The reframe also correctly signals that the fourth bullet is a manual walk-back rather than an automatic cleanup layer, which was the underlying concern.

The PR description was wrong — I claimed the launcher and design-doc corrections already existed at this PR's base, but KBR-78 reverted both during its merge. PR description now reflects what actually happened: this PR re-applies the corrections. Thanks for catching that; without the review this would have shipped a half-done "correction" of the same stale claim in three places.

PR #172, opened 2026-09-15
KBR-260: byte-exact backup restore across platforms (newline="")

shelpuk:

`kitty cleanup` is the repair command a user runs when a crashed `kitty claude` session leaves their `~/.claude/settings.json` damaged. Its promise is exact: put the user's file back **byte-for-byte** as it was before kitty touched it.

This PR fixes a silent violation of that promise. On Windows, a restored file came back with CRLF line endings where the original had LF — every line ending changed, invisibly, by kitty itself. The same class of defect also silently corrupted restores on Linux/macOS whenever the backup file carried CRLF line endings. After this fix, the restore is byte-exact on every platform, and the test that documents this promise now runs on Windows too instead of skipping.

No user-facing workflow changes; no configuration changes; the repair command behaves exactly as documented — it just finally keeps its word on Windows.

## What changed

Three function bodies, one test-skip removal, tests added to two existing files:

| Site | Change |
|---|---|
| `src/kitty/launchers/claude.py::_atomic_write_text` | `os.fdopen(..., newline="")` — disables CPython's `\n` → `os.linesep` translation on write (the Windows CRLF corruption) |
| `src/kitty/launchers/claude.py::load_settings_backup` | `path.open(newline="")` instead of `Path.read_text` — disables universal newlines, which strip `\r\n` → `\n` on read on **every** platform |
| `src/kitty/cli/cleanup_cmd.py::_load_backup` | same read-side fix, symmetric with the launcher's reader |
| `tests/cli/test_cli_settings_lifecycle.py` | removed `@pytest.mark.skipif(sys.platform == "win32")` from the L3 AC-4 byte-exact restore test — per TEST_SUITE.md §9, a skip on a false-on-Windows assertion is a defect with a ticket; KBR-260 is that ticket |
| `tests/cli/test_cleanup_cmd.py`, `tests/test_claude_settings_override.py` | L1 coverage: per-reader staged-bytes byte-identity tests (gate on every leg), a `run_cleanup` end-to-end CLF/CRLF restore test, an `os.fdopen`-kwargs contract test (the only POSIX-runnable write-side regression guard), and Windows-only behavioural tests CI-verified on the Fast gate's Windows leg (KBR-164 matrix) |

`_atomic_write_json` is deliberately untouched — its output is kitty-generated JSON where line-ending style is not part of any byte-identity contract.

## Why the read side too (deviation from the ticket's proposed fix)

The ticket proposes only the write-side `newline=""`. That alone does not satisfy the ticket's own contract ("restore byte-for-byte even when the backup is byte-identical to the pre-launch content"): the ticket's root-cause analysis claims `read_text(encoding="utf-8")` "honors the input as written", but `read_text` opens in universal-newlines mode and translates `\r\n` → `\n` in memory on every platform. With a write-only fix, a CRLF backup would restore as LF — wrong on both platforms. Empirically verified on POSIX during the requirements review; full analysis in `.requirements/20260915T163410Z_kbr_260_atomic_write_newline/REQUIREMENTS.md` §7.

## Testing evidence

- New L1 tests were written RED first and observed failing for the right reasons; all green after the fix.
- Mutation kill demonstrated for all three `newline=""` sites: removing each in turn makes the corresponding L1 test fail.
- Windows-only behavioural tests (`skipif != win32`) run on the Fast gate's Windows CI leg, where they are RED before this fix and GREEN after.
- Full suite: 6876 passed / 9 skipped locally; ruff, mypy (no new errors vs main), import-linter 5/5 contracts kept.


shelpuk:
**Sibling defect verified and filed as KBR-262.**

We verified the finding against `src/kitty/launchers/kilo.py`: `prepare_launch` reads the user's config with `Path.read_text(encoding="utf-8")` (`kilo.py:143`, universal newlines — strips `\r\n` on every platform) and `cleanup_launch` writes it back with `Path.write_text(original, encoding="utf-8")` (`kilo.py:194`, default newline — adds `os.linesep` on Windows). Same defect class as KBR-260; not fixed here because the scope of this PR is KBR-260, which names `_atomic_write_text` + the two backup readers + the Claude restore path, not the kilo adapter.

Filed as **[KBR-262](https://shelpuk.atlassian.net/browse/KBR-262)** with the same fix shape (`newline=""` on both sides, `Path.write_text(newline=...)` is available since Python 3.10), the same L1 test plan (per-leg staged-bytes + Windows-only behavioural on the Fast gate's Windows leg), and the same scope note leaving `kilo.py`'s `_atomic_write_json` alone. Not fixing it in this PR keeps the diff surgical and lets KBR-262 land with its own tests.

Thanks — the sibling sweep is exactly the right check; `launchers/` is the one place the same defect class would recur, and this was the only other user-content round-trip path.


[KBR-262]: https://shelpuk.atlassian.net/browse/KBR-262?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #175, opened 2026-09-15
KBR-261: write_entry emits LF-only manifest bytes on every platform

shelpuk:
## Context

The "kitty-bridge" golden corpus is the suite's evidence layer: real Claude Code `/v1/messages` bodies captured into `tests/corpus/<id>.body`, with provenance and `body_sha256` recorded alongside each in `<id>.json`. The corpus' byte-exactness guarantee depends on every committed artifact staying LF — every assertion that reads a body (M3 length, M5 budget, M13 compaction) and every test that reads the manifest through `load_corpus` assumes that.

`.gitattributes`'s `tests/corpus/*.body -text` / `*.json -text` rules (added KBR-29) close one side: git stops translating LF to CRLF on a Windows contributor's checkout. The other side — what the writer actually emits — was open. `corpus.write_entry` called `Path.write_text(json.dumps(...) + "\n", encoding="utf-8")` with no `newline=` argument, so on Windows Python's text-mode translation rewrites every `\n` to `\r\n` regardless of `.gitattributes`. A Windows regen run committed CRLF manifests, and CI's L1 roundtrip test reported the resulting LF-vs-committed byte diff as pure line-ending drift — the failure looked like a real regen problem, but the writer was the bug.

This PR closes that hole: the writer now passes `newline="\n"` so its bytes are LF on every platform, and the suite now has two tests pinning the guarantee end-to-end.

## Changes

- **`tests/harness/corpus.py::write_entry`** — one-line fix: `newline="\n"` on the `Path.write_text(...)` call. Platform-neutral widening, no behavior change on Linux.
- **`tests/harness/test_corpus.py::TestTheRoundTrip::test_the_manifest_bytes_are_lf_only`** *(L1)* — asserts `b"\r\n" not in path.read_bytes()` after `write_entry`. Reads bytes (not text) so universal-newlines translation cannot mask the drift. Weak on Linux (no bug to detect), strong on the `windows-latest` CI leg.
- **`tests/harness/test_corpus_lint.py::TestTheCorpusIsProtectedFromLineEndingTranslation::test_no_committed_manifest_carries_crlf`** *(L2)* — iterates every committed `.json` in `tests/corpus/` with `read_bytes()` and refuses any that contains `\r\n`. Counterpart to `test_gitattributes_still_covers_the_corpus`: that pins the checkout end, this pins the write end.
- **`.system_design/TEST_SUITE.md` §7.1.1** — records the two lines of defence (KBR-29's `.gitattributes` rule + this PR's `newline="\n"` fix) as the design rationale for line-ending protection, which §7.1.1 had not previously stated.
- **`tests/harness/test_corpus_thresholds.py`** — its roundtrip test's docstring documented the Windows CRLF behavior as the reason the comparison goes through `_manifest_for + dumps` rather than `write_entry`. That claim is no longer true; the docstring now reflects post-fix reality and points at the L2 guard.

## Out of scope

KBR-260 (sibling ticket) owns the same root cause at a different call site — `_atomic_write_text` in `src/kitty/launchers/claude.py`. Already merged via PR #172. Each fix is scoped to its own call site and its own test surface; this PR does not touch launcher code.

## Test plan

| Layer | Test | What it pins |
|---|---|---|
| L1 | `TestTheRoundTrip::test_the_manifest_bytes_are_lf_only` | `write_entry`'s own bytes carry no CRLF |
| L1 | `TestCommittedArtifactsRegenerate::test_every_committed_manifest_matches_write_entry_byte_identically` *(existing, docstring updated)* | Builder's manifest equals the committed bytes via writer-independent comparison |
| L2 | `TestTheCorpusIsProtectedFromLineEndingTranslation::test_no_committed_manifest_carries_crlf` | Committed tree has no CRLF in any `.json` |
| L2 | `TestTheCorpusIsProtectedFromLineEndingTranslation::test_gitattributes_still_covers_the_corpus` *(existing)* | `.gitattributes` patterns still name the corpus directory |

Local sweep: `pytest -m "l1 or l2"` (harness only) — 1812 passed. Full repo sweep — 4967 passed, 2 skipped.


shelpuk:
Claude Code review suggestion applied: the §7.1.1 heading now reads **"Three lines of defence keep the committed bytes LF"** — the paragraph describes the `.gitattributes` checkout-end rule, the manifest digest in `load_entry`, and the KBR-261 `newline="\n"` writer fix. Pushed as `a1c5167`. Thanks for catching the drift.

PR #176, opened 2026-09-15
T-F2 (KBR-71): compaction properties + shadowed context_window visibility

shelpuk:
## Context (Product-Owner)

T-F2 lands the four §6.1 **property tests** for `_compact_messages` — the I1-fidelity core of the bridge — together with the **shadow-log remedy** KBR-170 folded into this ticket. Together these close two distinct product promises:

1. **I1 Message Fidelity is now property-tested, not example-tested.** Every `_compact_messages` invocation is now checked against the four invariants the design names — identity below budget, no orphan or unanswered tool pair, idempotence, and the irreducible exception — across every conversation the T-F1 substrate generates (200 examples per shape, both Chat-Completions and Anthropic-native). A future regression in grouping, head/tail accounting, or the fallback loop is caught at the property, not in production.

2. **KBR-170's silent shadowing is now loud.** The owner kept the overrides-catalog on top of `provider_config["context_window"]`, but the operator who set that window by hand could not discover they were being outvoted by a file synced from the network. `get_model_context_tokens` now announces each shadowed lookup at `INFO` once per `(provider, model, effective, ignored)`, naming the model, both values, and the remedy. **No behaviour change to the precedence** — only visibility.

The PR also closes the §6.1 "Open, referred onward" note in `TEST_SUITE.md` and updates `get_model_context_tokens`' docstring to match.

## What ships

- **`src/kitty/providers/model_context.py`**: new `_log_shadowed_context_window` (deduped via `@functools.cache`); wired into `get_model_context_tokens` after `_lookup_override` succeeds and only when the two values disagree; docstring updated to state the precedence and the log.
- **`tests/test_model_context.py`**: new `TestShadowedContextWindowIsObservable` (7 cases) — fires on disagreement, silent on every other branch of the truth table (matching value, missing profile value, missing override, uncoercible profile value, second-lookup silence, re-announce on catalog change). Existing `test_overrides_still_outrank_provider_config` unchanged.
- **`tests/bridge/test_compaction_properties.py`** (new, 11 tests): the four properties plus the no-raise pin, both shapes, with `_StubLauncher`/`_StubProvider` mirroring the existing test-compaction pattern.
- **`.system_design/TEST_SUITE.md`**: §6.1's "Open, referred onward" paragraph replaced with the owner decision (2026-09-15); no other KBR-170 references remain open.

## Property-test falsification (plan §1.4)

Each property was watched failing against a deliberately broken compactor:
- **P1** killed by a "drop everything" stub.
- **P2** (sharpened, bidirectional) killed by a **non-atomic-grouping** compactor on the native shape — `TestNoOrphanedPair::test_messages_output_is_pairing_clean`, `TestIdempotence::test_messages_second_compaction_is_a_no_op`, `TestBudgetOrIrreducibleOrRaises::test_messages_valid_input_never_raises_compaction_failed` all failed under the mutant across the 200-example sweep.
- The one-directional P2 form (results ⊆ uses only) **survived** the same mutant — review finding **B1** — confirming the bidirectional oracle is the form that catches the grouping bug. The old form would have left the mutation in place.

## What's out of scope

- **Q10** (what *should* happen when the final turn alone will not fit) — deferred per the ticket; the property lands the observed-behaviour form and is revised once Q10 is answered, together with TR-3 and register rows M3–M7/M13.
- **Mutation-threshold wiring (T-H3)** consumes these properties; it does not extend them.
- **Changing the precedence** — owner decision 2026-09-15: overrides catalog stays on top.

## Gates

- `pytest -m l1 -q`: **OOM-killed locally** under concurrent sessions (per repo memory "killed ≠ failed; CI is authority"). Targeted runs (169 tests across the touched files) are green; CI's runner will be the deciding gate.
- `ruff check .` on touched files: **clean**.
- `lint-imports`: **5 contracts kept**.
- `mypy src/kitty tests/bridge/test_compaction_properties.py tests/test_model_context.py`: **Success, no issues found**.
- Layer guards (`tests/test_layer_markers.py`, `tests/test_layer_selection.py`): **70 passed**; the new property file takes the `l1` path default and needs no registration.


shelpuk:
Both findings addressed — the warning with a code change (36205a8), the suggestion with a clarification, no change needed.

**Warning — operator-visibility claim contradicted the logging setup.** Agreed and fixed. "Without enabling debug mode" was true in the narrow sense (the line is at INFO, not DEBUG) but overstated: whether INFO is visible at all is the operator's logging configuration, and a handler that filters INFO out makes the line invisible — the exact failure mode KBR-170's docstring escape hatch already suffered from, in a different form. 36205a8 qualifies both surfaces (the `get_model_context_tokens` docstring and §6.1's decision paragraph): the shadowing is discoverable **at INFO level, dependent on the operator's logging config emitting at INFO for `kitty.providers.model_context`**, with handlers that filter INFO called out explicitly. The INFO-not-WARNING rationale (precedence is deliberate — visibility, not an alarm) is unchanged.

**Suggestion — deadline on the property tests (low confidence).** No change: the constructed oversized case (`test_constructed_oversized_final_turn_lands_in_the_irreducible_arm`) is a plain `def test_` — not `@given`-wrapped — so hypothesis's 200 ms per-example deadline does not apply to it; it runs once, deterministically. The ten `@given` properties generate bodies from the T-F1 substrate's short-ASCII strategies (≤ 16 chars per text draw, ≤ 3 tool declarations, ≤ 13 messages), so a serialized body is a few KB and one compaction pass is ~1 ms — two orders of magnitude under the deadline. Under `CI=1` the profile registered in `tests/conftest.py` removes the deadline and derandomises anyway (the T-F1 pattern).

Both test files re-run green after the docstring change (90 passed); ruff clean on touched files.

shelpuk:
The orphaned finding is addressed in `46a0d96` — with the code remedy, not just the README one, because on verification the defect is one layer deeper than the finding stated.

**Verification.** `_setup_debug_logging` attached its FileHandler only to `kitty.bridge` (confirmed at server.py:2723/2741). But a handler alone would not have been enough: **nobody sets a level on `kitty.providers.model_context`**, so its effective level came from root's default WARNING, and the INFO records were dropped at the logger level before any handler was consulted. The shadow notice was undeliverable via `--debug`, and so — with the same defect since it was added — was KBR-151's default-fallback line.

**Fix.** `_setup_debug_logging` now attaches the same file handler (same path, same format, deduped by the existing marker attribute) to `kitty.providers.model_context` as well and sets that logger to DEBUG inside the debug path. Without `--debug`, surfacing INFO remains the operator's logging configuration, as the qualified docstring from the previous round already states — but the product's own diagnostic path now delivers what the KBR-170 acceptance criterion promises.

**README.** The *Logging* section now names the budget-resolution lines and the `kitty.providers.model_context` logger, per the finding's alternative remedy.

**Test.** `tests/test_debug_log_path.py::TestBudgetResolutionLinesReachTheDebugLog` — red before the fix (file empty, exactly the operator-visible symptom), green after. It pins both notices reaching the file; the class carries an autouse handler-cleanup fixture mirroring the sibling class' `kitty.bridge` cleanup, because the dedup marker's per-logger shape (fine in production, where one process runs one server) needs isolation across tests.

184 tests green across the touched files; ruff clean; `test_readme_table_guards.py` green after the table edit.

shelpuk:
Round 3 acknowledged — no further changes. Both prior warnings confirmed fixed (36205a8 docstring qualification, 46a0d96 debug-path delivery + README), and the only remaining item is the low-confidence deadline suggestion, which this review itself notes follows the project convention (`tests/harness/test_transcripts.py` pattern; the constructed case is a plain `def test_`, so no per-example deadline applies). Leaving it as-is.

The review chain has converged: rounds 1–2 each produced a substantive defect that is now fixed and pinned by a test; round 3 produced none. CI on the amended commit is running; merge left to the owner.

shelpuk:
Round 4 acknowledged — no further changes. The only remaining item is the same low-confidence deadline suggestion carried forward three rounds; this round's sweep confirms the prior fixes (36205a8 docstring qualification, 46a0d96 debug-path delivery, 220cf06 vendor-token-guard classification) and adds no new finding.

Out-of-band note (not part of the review): the same commit also re-runs the test matrix after the vendor-token-guard classification — three Ubuntu test-matrix jobs had been failing on the previous commit because the new `logging.getLogger("kitty.providers.model_context")` call in server.py tripped `tests/bridge/test_vendor_token_guard.py`; classifying the literal as a logger name in `_ALLOWLIST` (same category as the existing `kitty.bridge` entry) is what the guard's design expects for new logging-tree identifiers and is what the KBR-71 `--debug` wiring makes honest.

PR #192, opened 2026-09-16
T-F3 (KBR-72): pairing and truncation properties + Responses substrate

shelpuk:

Kitty Bridge sits on the wire between a coding agent and an LLM provider. Two of its safety nets are (a) **tool-call pairing** — if a conversation carries a tool result whose declaring tool call is missing, the upstream provider rejects the whole turn with error 2013, so the bridge strips such orphans before they ship — and (b) **oversize truncation** — a single huge tool output can alone exceed a model's context window, so the bridge shrinks it to a short notice before sending. Until now those two rules were proven only by a handful of hand-written examples: tests that pick one input each and prove that input. This PR replaces that coverage gap with **property-based tests** — tests that state the rule ("no orphan ever survives"; "nothing under 50,000 characters is ever touched") and let the `hypothesis` library generate hundreds of random conversations trying to break it.

**Scope extension (owner decision 2026-09-16).** The design review found the Responses-format twins of these two rules (`_truncate_oversized_responses_outputs`, `_drop_orphan_responses_tool_outputs`, register rows M3/M7) had **no L1 property owner anywhere in the plan and no direct example-test caller anywhere in `tests/`** — they were only exercised indirectly through end-to-end runs. The owner chose to extend this ticket rather than file a separate one, so this PR closes that gap too: the shared transcript substrate gains OpenAI Responses strategies, and both twins get the same property coverage as their CC/native siblings.

## What landed

1. **Substrate extension** (`tests/harness/transcripts.py`): Responses-shape strategies (`responses_tool_definition`, `responses_function_call_item`, `responses_function_call_output_item`, `responses_message_item`, `responses_request`) + `responses_problems` validity reporter, following the per-format conventions the KBR-70 substrate established. Pinned by 14 new tests in `tests/harness/test_transcripts.py` (export surface, 200-example generated validity, JSON-strictness, pairing invariant, reporter falsification ×8 incl. forward-reference and role-alternation guards, innocent-text negative control, NaN rejection, hostile-input rows). A stale forward-reference comment about `messages_tool_pair` was corrected while there.
2. **The property file** (`tests/bridge/test_pairing_truncation_properties.py`, 40 tests): §6.1 pairing and truncation properties over all three wire shapes.
   - *Pairing*: output contains no orphan in either message shape (CC `tool` messages; native `tool_result` blocks; Responses `function_call_output` items), with paired results and all other content surviving structurally identical and in order — asserted by equality against a wire-rule oracle, so a drop-everything mutant cannot pass.
   - *Truncation*: identity below the 50,000-char limit (boundary 50,000 untouched; exact-50,001 truncated, all three shapes), bounded above it (exact notice text pinned; the return count checked against an independent pre-snapshot oracle so a miscounting mutant cannot agree with itself), non-tool content untouched (oversize non-tool text, structured list `tool_result` content, list-form Responses outputs).
   - *Constructed cases*: boundary 50,000/50,001, falsy-id orphans (`tool_call_id=""` etc.), forward-references (output before its declaring call — documents the preceding-id reading).
   - *§1.4 harness rule*: eleven falsification cases proving every oracle rejects a deliberately wrong output, each with a hand-written expected so an oracle defect cannot agree with itself through the generated property and the constructed case.
3. **Negative control (out-of-tree)**: an off-by-one mutation of the truncation boundary predicate (`>` → `>=`, scoped to `_truncate_oversized_tool_results`) was run against a scratch copy — exactly the two boundary tests it should kill died; the live tree was never touched.

## Review + gates

- REQUIREMENTS reviewed by system-design-reviewer (3 blockers + 6 concerns + 7 suggestions — all applied: index-shift fix, twins scope decision, import-discipline fix, count-from-pre-snapshot, per-run vs per-example arm coverage, added drop-direction falsification, em-dash/line-number/terminology cleanups).
- Code reviewed by code-reviewer (REQUEST CHANGES → all findings applied: ~37 missing Google docstrings added, stub moved to module level per sibling convention, two untested Responses reporter guards pinned, duplicated docstring paragraph collapsed, RST backtick fixed, duplicate native tests split by arm, inline tool literals replaced with helpers). Re-review gates green.
- Local gates: `ruff` clean · `mypy src/kitty` clean (94 files) · `lint-imports` 5/0 · layer guards 70/70 · target files 113 passed / 2 skipped · full L1 suite green (6,441 passed / 0 failed) on the pre-review-fix commit; re-run against this final commit in progress, CI is the authority for the full matrix.
- Requirements doc: `.requirements/20260916T122734Z_t_f3_pairing_truncation_properties/REQUIREMENTS.md` (local, gitignored per repo convention).

Closes KBR-72 (T-F3, Epic F).


shelpuk:
Fixing the failing macOS leg and responding to the automated review's suggestions.

**Why the macOS leg failed (not a T-F3 regression).** The `test / Python 3.12 on macos-latest` job failed in `ruff check .` with 5 SIM117 findings in `tests/bridge/test_stage6_http_correctness.py` — a file this PR never touches. Root cause: the repo's dev pin is unbounded (`ruff>=0.4`), and the macOS runner resolves a newer ruff than the Linux / Windows runners. The same findings were caught and auto-fixed on `main` today by a sibling session (commit `9cb2f24`, PR `#194`, "lint: keep ruff 0.16.8's SIM117 finding out of the gate") — my branch was cut from `main` before that fix landed, so the pre-existing drift reappears here.

**Fix applied.** Merged `origin/main` into this branch (no-ff, merge commit `aad567f`) to absorb the SIM117 fix rather than re-applying an identical change. After merge: `ruff` clean, `mypy src/kitty` clean, layer guards 70/70, target files 113 → 115 passed / 2 skipped. Pushed as commit `01f2a4c`.

**On the automated review's three suggestions:**

1. *Local `_short_text` duplicates the substrate's `_short_ascii_text` (deliberate)* — agreed, and documented in the property file's header: the property tests need BOTH short content (orphan injection) and long content (truncation over the 50,000 limit); the substrate's helper caps at 16 chars, so it is unusable for the truncation property. Keeping the local copy also preserves the substrate's "downstream consumers import no private names" posture.

2. *"responses_problems does not assert output type shape"* — **real coverage gap, addressed in commit `01f2a4c`.** The published schema allows `function_call_output.output` as a string OR a list of `input_text`/`input_image`/`input_file` parts; the reporter now rejects a non-string non-list value and shape-checks the per-part contents on the list branch. Two falsification rows added to `_RESPONSES_DEFECTS` proving the new branches fire.

3. *"reasoning item type is reporter-allowed but strategy-uncovered"* — observation accepted, no strategy added. `reasoning` is an opaque item type at the wire-projection reader level (T-A3's reader explicitly names opaque item internals as a deliberate blind spot), so teaching the substrate to emit opaque reasoning payloads would couple the substrate to a shape the reporters do not pin. The response-direction strategy coverage stays with the modelled, wire-rule-relevant item kinds (`message`, `function_call`, `function_call_output`); if a future register row makes reasoning items semantically relevant at L1, the strategy set grows then. Recorded here so the symmetry gap is a decision, not an oversight.

CI is re-running on `01f2a4c` (all eight test legs + review + Analyze + CodeQL + ci-required + update-metadata). Will address further review rounds if the reviewer produces more findings.

PR #196, opened 2026-09-16
KBR-221: carry tool_choice and parallel_tool_calls on the Responses and Gemini ingresses

shelpuk:
# KBR-221 — ResponsesTranslator and GeminiTranslator silently drop `tool_choice`

**Fixes** [KBR-221](https://shelpuk.atlassian.net/browse/KBR-221) — sibling of KBR-214 (merged as PR #100), which closed the same defect on the Anthropic Messages ingress.

Claude Code (and other agent CLIs) can send a *tool choice* with a request — the instruction that says whether the model must call a tool, must not, or must call one specific tool. On kitty's translated Anthropic-Messages route this instruction has reached the backend since KBR-214. On the **Responses** route (Codex CLI) and the **Gemini** route (Gemini CLI) it did not: the two ingress translators rebuilt the request for the Chat Completions backbone and silently dropped the field, so a *forced* tool call became *optional* upstream and a *forbidden* one became *possible* — with no error and no log line. Same user impact as KBR-214 had on Messages.

This PR fixes both routes by mapping the instruction onto the form the backbone already speaks, reusing the seams KBR-214 built. Every backend the bridge can reach now receives the instruction it can express:

- **Responses route** — `tool_choice` strings (`auto`/`none`/`required`) and named-function choices are carried; `parallel_tool_calls: false` (the parallel-tool-use knob) is carried. Shapes with no backbone equivalent (hosted tools like `web_search_preview`, MCP, custom tools) are deliberately omitted — forcing a call to a tool nothing on the route can execute would be worse than not forcing one — and the omission is now documented.
- **Gemini route** — `functionCallingConfig.mode` (`AUTO`/`ANY`/`NONE`, case-insensitive) is carried; `ANY` with a single allowed function maps onto the backbone's named form. Shapes the backbone cannot express (a *set* of allowed functions, Google's `VALIDATED` mode) carry the mode only — the owner chose "carry the representable half, record the loss" over dropping the whole instruction.

Per KBR-214's precedent, omissions are recorded rather than left silent: the PR adds a §3.3.2 prose claim in `TEST_SUITE.md` naming every shape each carrier omits, with the rule that a register row becomes due when a corpus entry first carries one.

**Out of scope, filed separately:** the harness's own Responses reader writes an explicit `parallel_tool_calls: true` verbatim while the Chat Completions reader suppresses it per §3.3.1b — an oracle-level asymmetry, not a product defect; noted in §5 of the requirements and flagged for a follow-up ticket.

## What changed

| File | Change |
|---|---|
| `src/kitty/bridge/responses/translator.py` | `carry_responses_tool_choice(responses_request, cc_request)` + `_names_degraded_responses_tool` helper; wired into `ResponsesTranslator.translate_request` |
| `src/kitty/bridge/gemini/translator.py` | `carry_gemini_tool_choice(gemini_request, cc_request)`; wired into `GeminiTranslator.translate_request` |
| `tests/bridge/test_responses_translator.py` | `TestResponsesToolChoice` — 44 cases: value table, D5/D8/D9/D10/D2 omissions, cross-product composition |
| `tests/test_gemini_translator.py` | `TestGeminiToolChoice` — 50 cases: case-insensitive mode table, names-loss policy, D9 gate, composition |
| `.system_design/TEST_SUITE.md` | §3.3.2 prose claim for the carriers' irreducible omissions (KBR-139 precedent) |

## Design decisions worth calling out

- **One carrier per ingress, not a shared table.** The source shapes differ per wire (Responses publishes nine choice forms; Gemini publishes mode + names); the destinations share one. Mirrors KBR-214's `carry_tool_choice_and_metadata` placement.
- **The "no tools, no choice" gate (D9) reads the CC list, not the inbound list.** A Responses `tools` list of only hosted entries filters to an *empty* CC list — gating on the inbound key would carry a choice the backbone 400s on.
- **The named-tool lookup (D10) walks the inbound list** — the only place a degraded tool still lives (the CC list has already filtered it out). Undeclared names are *carried*: the agent's mistake, and the provider's 400 names it (D8).
- **No register row in this PR.** The fix carries a field the bridge used to drop, so no delta exists to claim; the KBR-184 rows P31/P32/P33 already account for the destinations that still cannot carry it.

## Verification

- **Red at base** (`origin/main` `4f38421`): 26 carrier tests fail with `KeyError: 'tool_choice'` / `'parallel_tool_calls'` — the exact defect; 35 negative-control cases (the omit behaviours) pass at base and stay green.
- **Green with the fix**: full local gate — `ruff check .`, `mypy src/kitty`, `lint-imports`, layer guards, `pytest -m "l1 or l2"`: **7490 passed, 13 skipped**.
- **KBR-214 non-regression** (AC-13): `TestToolChoiceAndMetadata` (33 cases) green; the destination seams (`_anthropic_tool_choice`, `_converse_tool_choice`, `_responses_tool_choice`) verified by direct execution against every value the new carriers emit — including Bedrock's P33 `auto`-rewrite for absent/`none`.
- The branch was rebased onto `origin/main` `f61a886` mid-flight (KBR-263/266 landed); the §3.3.2 prose claim coexists with the KBR-263 M16 amendment without conflict.

[KBR-221]: https://shelpuk.atlassian.net/browse/KBR-221?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


shelpuk:
@github-actions[bot] addressed in 8df899e — dropped the dead constant and reworded the docstring to state the property the helper tests (`type != "function"`) rather than enumerating the closed set in two places. The eleven hosted types stay documented in TEST_SUITE.md §3.3.2 prose and in the test_tool_choice_hosted_is_omitted parametrize ids (tests that kill a constant-removal mutant). All local gates green: ruff, mypy, lint-imports, 368/368 tests.

shelpuk:
**Windows leg failure — diagnosed as the known pre-existing flake, not this PR's regression. Rerunning the failed leg.**

`TestTheWindowsConsoleDetachment::test_a_background_bridge_survives_a_console_break_in_its_launching_console` failed at `tests/bridge/test_bridge_management.py:1909` with `json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)` — `json.loads(snapshot_path.read_text())` read an **empty** file whose `exists()` check had just passed: the child created the snapshot file but the parent read it before the write flushed. This is the create-then-write race this repo has seen on this exact test before (no code relationship to KBR-221: the branch's diff touches `bridge/responses/translator.py`, `bridge/gemini/translator.py`, the two translator test files, and `TEST_SUITE.md` §3.3.2 — zero lines in `test_bridge_management.py` or any launcher/console code; verified with `git diff origin/main..HEAD --stat`).

The other five test legs — ubuntu 3.10/3.11/3.12/3.13 and macOS 3.12 — all passed on this same commit. Failed leg rerun via `gh run rerun --failed`; will report the outcome here.

PR #197, opened 2026-09-16
KBR-67: AST-level start-path domination guard

shelpuk:
## Context for Product Owner

KBR-67 closes **gap #3** in `TEST_SUITE.md` §5.1: the existing start-path coverage guard was file-granular — it asked only whether a file contained *both* a `BridgeServer(` and an `egress_block_reason(` call. `cli/main.py` already holds **two** `BridgeServer(` start paths today, so a third start path added to that same file tomorrow would have passed the old guard unguarded and become an egress leak — exactly the failure mode invariant **I3** (egress containment) exists to prevent.

This change tightens that guard to **AST-level domination**: every `BridgeServer(` construction in `src/kitty/**.py` (outside the class definition) must be lexically preceded by an `egress_block_reason(` call **in the same function scope**, not just the same file. The five live start paths — `bridge_runner.py` ×2, `cli/main.py` ×2, `cli/launcher.py` — are all already correctly ordered, so this lands as a hardening of the test suite rather than a runtime fix. The next person who adds a start path to `cli/main.py` gets caught by the guard at PR time.

Per `TEST_SUITE.md` §1.4 (the harness rule), the guard carries its own **falsification control** running in the suite — an in-test source fixture with one undominated and one dominated `BridgeServer(` proves the AST walker can detect a violation. Without it, a broken AST walk could silently turn the guard into a no-op while every test still passes.

## What landed

- **`tests/test_egress_coverage.py::TestEveryStartPathIsGuarded`** replaced with an AST-level version:
  - `test_every_construction_is_dominated_by_a_guard_call` — AC1, AC2.
  - `test_the_scan_finds_the_known_start_paths` — self-guard for known positives (AC4).
  - `test_undominated_construction_is_caught` — falsification control (AC3).
  - Module-level helpers `_iter_bridge_constructions`, `_enclosing_function`, `_is_dominated`, `_called_name` with Google-style docstrings; the existing file-granular helpers (`_files_calling`) are removed.

- **`.system_design/steps/ast_start_path_domination_guard.md`** — step file for traceability, with implementation notes and a status line.

## Verification

- `ruff check` clean.
- `lint-imports` clean (5 contracts kept).
- `mypy src/kitty` clean.
- `pytest -m "l1 or l2"` — 7,668 passed, 13 skipped (platform-bound only), 0 failed.
- Empirical falsification (done during development, then reverted): appending an unguarded `_falsification_unguarded()` function to `bridge_runner.py` produced `AssertionError: ... bridge_runner.py:230 in _falsification_unguarded()`. The guard caught it.

## Out of scope (by KBR-67's own definition)

- **T-E6 (KBR-66)** — the *behavioural* enforcement test: that the guard's rejection actually prevents startup (drives five start paths with a rejecting config and asserts no server starts, with its own falsification). That ticket is KBR-67's explicit complement.
- No source-code changes — this is a test-only change. The five live start paths were already correctly ordered; the ticket hardens the guard so a future addition stays that way.

## Spec

- KBR-67 (Jira).
- TEST_SUITE.md §5.1 gap 3, §5.4 row "Start-path domination", §6.2.3 row "Start-path domination", §1.4 (harness rule).
- Task requirements: `.requirements/20260916T204727Z_ast_start_path_domination_guard/REQUIREMENTS.md`.


shelpuk:
Thanks — pushed a refresh in commit `5b8d574` ("KBR-67: refresh TEST_SUITE.md passages this guard obsoletes") addressing all three:

- **§5.1 gap 3** — updated to describe the guard as AST-level now and reference KBR-67 / the test file.
- **§5.4 row** — "**Must be strengthened**" replaced with "**Landed with KBR-67** as `tests/test_egress_coverage.py::TestEveryStartPathIsGuarded` (…§5.1 gap 3, now closed)", following the pattern of adjacent rows.
- **§6.2.3 Start-path domination row** — appended the landed-with note and a brief description of the test's three shapes (sibling-undominated, sibling-dominated, inner-function), matching the depth other landed rows in the table carry.

The previous two review-rounds addressed: a third case in the falsification fixture (guard-in-outer + construction-in-inner) to close the innermost-vs-outermost blind spot, and a redundant `Args:` block on `_iter_bridge_constructions` (commit `bffebfa`).

shelpuk:
Addressed the residual §5.1 heading in commit `43c6f82`: changed "Three things are missing, and they are the gap" to "Two things are missing, and they are the gap — a third (the file-granular start-path guard) has since been closed, and is recorded below in the state it reached."

Only items 1 and 2 (`BridgeServer`'s serving path untested; no negative assertion anywhere) are gaps now; item 3 is kept in the list as a closed item per the table's own convention so the gap-to-landed transition stays traceable.

shelpuk:
Closed both round-3 findings in commit `eb759d0`:

**Finding 1 — match scope narrower than the regex it replaced.** `_called_name` now matches `Attribute.attr` as well as `Name.id`, so `mod.BridgeServer(...)` is no longer invisible. A new `test_dotted_construction_spellings_are_matched` proves the widening is load-bearing (parses a synthetic `kitty.bridge.server.BridgeServer(...)` and asserts the walker picks it up; reverting the Attribute branch makes the test fail with `assert 0 == 1`).

**Finding 2 — `_is_dominated` count a nested-deferred guard as dominating the enclosing scope.** The new `_calls_in_own_scope` walker skips `FunctionDef`/`AsyncFunctionDef`/`ClassDef`/`Lambda` subtrees of the enclosing function (their bodies are deferred and would not have run unless explicitly invoked) while keeping comprehensions (whose implicit scope executes eagerly). The falsification fixture gains a fourth shape — outer-construction with the only guard confined to a nested helper scope — and that case fails the old `ast.walk(func)` walker with `assert not True` (proven by reverting `_calls_in_own_scope` back to `ast.walk`).

**Aliased imports.** `from kitty.bridge.server import BridgeServer as BS` would hide a construction site from the matcher without import resolution. A separate `test_no_aliased_bridge_server_imports` makes any such import a guard failure, so a future rename of the canonical import forces a deliberate decision. All four BridgeServer imports in `src/` today use the canonical binding.

**Doc refresh in the same commit.** `TEST_SUITE.md` §6.2.3 Start-path domination row now notes the Attribute widening, the alias guard, and the "own scope, not nested" domination rule; the step file describes the four falsification shapes; `REQUIREMENTS.md` R3, AC3 and the Testing Plan table are amended to record the round-3 widening.

**Falsification trace (committed code is load-bearing).**
- Revert scope walk to plain `ast.walk(func)` → `test_undominated_construction_is_caught` fails on the fourth shape (`assert not True` at line 422).
- Drop the Attribute branch from `_called_name` → `test_dotted_construction_spellings_are_matched` fails with `assert 0 == 1`.

shelpuk:
Closed both round-4 suggestions in commit `9062b11`:

**Suggestion 1 — generator expressions are lazy.** `_DEFERRED_SCOPES` now includes `ast.GeneratorExp`; the comment above the constant is rewritten to call out the asymmetry (ListComp/SetComp/DictComp stay eager, GeneratorExp does not). Falsification fixture gains a fifth shape — outer-construction with the only guard inside a generator expression — and a regression that drops `GeneratorExp` from `_DEFERRED_SCOPES` makes `test_undominated_construction_is_caught` fail on that case.

**Suggestion 2 — alias-guard had no §1.4 falsification.** Added `test_aliased_bridge_server_import_is_detected`: parses a synthetic `from kitty.bridge.server import BridgeServer as BS` and asserts the matcher surfaces `line 1: \`BridgeServer as BS\`` as the offender. A regression like `endswith("BS")` (silent matcher drift) makes the new test fail with `assert [...] == ["line 1: \`BridgeServer as BS\`"]`.

Falsification trace (committed code is load-bearing):
- Drop `GeneratorExp` from `_DEFERRED_SCOPES` → `test_undominated_construction_is_caught` fails on the fifth shape.
- Change `alias.name.endswith("BridgeServer")` to `endswith("BS")` → `test_aliased_bridge_server_import_is_detected` fails.

Test count in `TestEveryStartPathIsGuarded`: 6 → **8** (round-4 added the generator falsification shape and the alias-match probe). Whole `test_egress_coverage.py` is now **18 tests**.

shelpuk:
Closed both round-5 inline findings in commit `d07046d`:

**Finding 1 — "four shapes" wording in three places while the fixture holds five.** All three updated:
- `test_undominated_construction_is_caught` docstring now opens "The five shapes exercise the pairwise combinations of {construction, guard} across {own scope, nested scope}, plus generator laziness" and lists five bullets (sibling-undominated, sibling-dominated, outer-guard/inner-construction, outer-construction with nested-helper guard, outer-construction with guard inside a generator expression).
- `step file` rewritten to enumerate all five.
- `TEST_SUITE.md` §6.2.3 "Start-path domination" row updated to "five shapes" with the same enumeration.
- `REQUIREMENTS.md` AC3 and Testing Plan table updated to match.

**Finding 2 — `test_aliased_bridge_server_import_is_detected` re-implemented the alias-match loop inline.** Extracted two shared helpers, now called by both the live-source scans and the synthetic-tree probes:
- `_iter_constructions_in_tree(tree) → list[ast.Call]` — used by `_iter_bridge_constructions` (live scan) and `test_dotted_construction_spellings_are_matched` (synthetic probe).
- `_iter_aliased_bridge_server_imports_in_tree(tree) → list[tuple[int, str, str]]` — used by `test_no_aliased_bridge_server_imports` (live scan) and `test_aliased_bridge_server_import_is_detected` (synthetic probe).

The same inline-copy issue affected the dotted-call probe; the extraction applies symmetrically.

**Falsification trace (committed code is load-bearing).**
- Mutation A (`endswith("BS")` in the shared alias helper) → `test_aliased_bridge_server_import_is_detected` fails. `test_no_aliased_bridge_server_imports` continues to pass because the live tree has zero aliases either way — exactly why the synthetic probe exists.
- Mutation B (drop the Attribute branch from `_called_name`) → `test_dotted_construction_spellings_are_matched` fails.

Test count unchanged: **8 tests** in `TestEveryStartPathIsGuarded` (18 in the file). Whole-file count: 18.

shelpuk:
Closed the round-6 inline finding (definition-file exclusion is broader than necessary) in commit `ee63e2e`:

**The exclusion's real shape.** The reviewer is right that the `class BridgeServer:` line itself never needed the file-level skip — a class declaration is an `ast.ClassDef`, which `_iter_constructions_in_tree` (a `Call`-only matcher) never sees. The exclusion is load-bearing only for a `BridgeServer(` **call** that the file might come to contain, and until now such a call would have been silently hidden.

**Two changes:**

1. **`test_no_bridge_server_construction_in_definition_file`** — asserts `bridge/server.py` holds zero `BridgeServer(` calls. A future factory or test helper added there fails this test and forces the decision (move it out, or handle it explicitly), instead of bypassing the egress-guard check silently. This is what makes the file-level skip sound rather than merely convenient.

2. **The constant's comment** now states the accurate reason: the class line needs no exclusion (wrong node type), but the call-shape skip does — and it points at the new self-guard as the mechanism that keeps the skip honest, replacing the old "every other `BridgeServer(` in src/kitty is a construction site" claim.

**Falsification trace.** Injected a `_review_probe` classmethod containing a `BridgeServer(...)` call inside `bridge/server.py` → `test_no_bridge_server_construction_in_definition_file` failed with the injected line number. Reverted; 19 tests green.

Test count: **9 tests** in `TestEveryStartPathIsGuarded` (19 in the file).

shelpuk:
Closed the round-7 finding (G2 gap-register row still described the guard as file-granular) in commit `f3691a1`:

- **`TEST_SUITE.md:5376` (§9.2 G2 row)** — rewritten to the register's own partial-closure convention (the G10/G6 form): "**Partly closed (KBR-67).** The start-path guard is no longer file-granular: the AST-level domination guard landed as `tests/test_egress_coverage.py::TestEveryStartPathIsGuarded` (§6.2.3). **Still open:** no-bypass unproven **for the bridge's serving path**; no negative assertion." The remedy cell marks "AST start-path guard landed (KBR-67)" so the row reads consistently with the three passages (§5.1, §5.4, §6.2.3) already updated in earlier rounds.
- **Step-file status line** — "PR pending" was past its use-by; updated to the current state.

This closes the last cross-reference the earlier rounds missed. Thanks for carrying the finding forward — it was raised as an orphan in round 1 and a warning in round 5 and the thread was closed without a readable answer both times.

PR #200, opened 2026-09-16
KBR-65: provider-aiohttp containment slice — T-E5

shelpuk:
## Context (for Product Owner)

This change closes the third of four per-transport gaps in the egress-containment invariant (I3, TEST_SUITE.md §5.5). The invariant's load-bearing claim is: when an egress gateway is configured, no provider-bound traffic from the agent or the bridge reaches the upstream except through that gateway. Three transport paths were proven before this PR (the bridge's own aiohttp session, KBR-62; the curl_cffi session used by the OpenAI subscription adapter, KBR-63). This PR proves the remaining **two aiohttp paths that bypass `_session_for`** — `ollama_cloud`'s own serving session and the `openai_subscription` OAuth login leg — and records a `proven` verdict for the `provider_aiohttp` row in the capability report.

**The OAuth login leg is the disproportionate one.** It runs at startup, before anything else has been proven, and is the path most likely to fire on a fresh machine (§5.5). The slice proves the leg routes through the gateway when egress is configured, and that an injected bypass (e.g. a future product rewrite that builds the OAuth session without `aiohttp_session_kwargs()`) is detected by the harness.

## What was implemented

- **New `tests/harness/test_provider_aiohttp_containment_slice.py`** — 14 tests, 9 gated. The slice covers both aiohttp paths in one file because the capability report has one row for them. The verdict gate tracks nine names: the five §5.2.2 phase tests on the serving leg plus four OAuth-leg phases.
- **`ProviderAiohttpContainment`** registered under the containment registry. `direct_route` is the resolver hook the protocol names for this slice — a self-contained `try/finally` patch on `socket.getaddrinfo` (T-E3's precedent).
- **`ProviderTlsRecordingUpstream`** in `tests/harness/provider_recorder.py` — the provider recorder speaking TLS, taken at construction because `SealedNetwork`'s factory path calls `start()` with no arguments.
- **Conftest gate extended to three slices**: `_record_proven_with_sibling_guard` takes a sequence of sibling rows; `_floor_unsupported_shape` accepts multiple direct-leg names (T-E5 has two — the serving phase 1 plus the OAuth direct phase). T-E2 and T-E3 behaviour is preserved.
- **Nine new gate unit tests** in `test_containment.py` pin the T-E5 floor shape, the writer, and the generalised sibling guard (including an end-to-end `PROVEN` writer test).

**No product (`src/kitty`) changes** — the existing seams (`provider_config['base_url']`, `OAUTH_TOKEN_URL`, `aiohttp_session_kwargs` module bindings) were sufficient, as they were for T-E3. The structural pin that `run_oauth_flow` uses `aiohttp_session_kwargs()` (tests/test_egress_coverage.py) is unchanged.

## Test plan

- `pytest tests/harness/` → 2323 pass, 2 skipped, 3 deselected (no regressions).
- `pytest tests/harness/test_provider_aiohttp_containment_slice.py` → 14/14 pass (on Python 3.13).
- The verdict gate's `PROVEN` writer is unit-tested end-to-end (finaliser is invoked with synthetic gate state; report reads `PROVEN`).
- The floor-shape writer is unit-tested on the two-direct-leg shape.

## Jira

- Closes KBR-65 (T-E5).
- Documents inheritance from T-E2 (KBR-62) and T-E3 (KBR-63) — the verdict-gate pattern, the Python ≥3.11 skipif floor (bpo-44011, confirmed against aio-libs/aiohttp#6044 on 2026-09-16), and the resolver-hook + module-binding discipline (the patch site must be the binding the product resolves).


shelpuk (commented):
Thanks for the review. Both suggestions applied in commit 1a4f111:

1. **Redundant local import of openai_oauth in _drive_oauth** — dropped the local `from kitty.auth import openai_oauth as _oauth`; the module-level import already binds the same name. The load-bearing-binding rationale (the drive must resolve the binding the phase-3 patch lands on) moved into the docstring above the construction line so the next reader sees why this construction line is what it is.

2. **TestThe-prefixed class name** — renamed `TestTheOAuthLoginLegContainment` → `TestOAuthLoginLegContainment`. Gate keys on method names, so this is internal-only.

Slice re-runs 14/14; lint clean. Watching CI for the new push.

PR #201, opened 2026-09-16
KBR-68: T-E8 — local bypass, fail-closed, and the transport asymmetry

shelpuk:
## Context

When the user configures an egress gateway in Kitty Bridge, three containment guarantees have to hold, and they had uneven test coverage: a local provider (a local Ollama) must connect directly without tunnelling because a rented proxy cannot reach the caller's LAN; a profile whose transport cannot honour the proxy (Bedrock in SSO mode) must refuse to start and tell the user which profile is the problem rather than silently leaking their real IP; and the proxy must be applied *unconditionally* on every outbound path that isn't the bridge's own aiohttp session — the design is deliberately asymmetric, and a future "fix" must not quietly add a local-bypass to those paths.

This PR delivers KBR-68 / T-E8: the test code that proves those three guarantees. No product behaviour changes; the bridge already implements all three. What's new is the proof, pinned at the cheapest layer that can prove each claim.

## Summary

Three test claims, each at the cheapest layer that proves it:

**R1 — Local bypass, end-to-end** (`tests/harness/test_local_bypass_slice.py`)
A real `BridgeServer` with the proxy gateway configured, pointed at a plain-HTTP loopback recorder on `localhost` and `127.0.0.1` (both branches of `should_bypass`), delivers directly with zero CONNECT attempts recorded by the harness proxy. The recorder asserts exactly one request with the expected body echo. Plain HTTP on loopback is the "local Ollama is plain HTTP" scenario — the delivered `SealedNetwork` recorder binds TLS, so a plain listener matches the shape.

**R2 — Fail-closed on a real start path** (`tests/test_egress_start_path.py`)
The bridge-runner single-profile start path (`src/kitty/bridge_runner.py:179-183`) refuses to start for a Bedrock-SSO profile under egress, names the profile in stderr, never binds a `BridgeServer` listening socket. A compliant non-SSO AWS key sails through (control). The falsification exercises the exact defect `TEST_SUITE.md` §6.2.3 names: the `egress_block_reason` call is preserved but its return value is discarded — the assertion surface proves the test is sensitive to that mutation.

> T-E8's R2 covers one start path. T-E6 owns per-entry-point coverage across the other four call sites (`bridge_runner.py` ×2, `cli/launcher.py`, `cli/main.py` ×2) and the only ticket that discharges §6.2.3's behavioural row. EG-3 (L4 Gherkin) remains Epic J.

**R3 — Transport asymmetry, pinned per site** (`tests/test_egress_asymmetry.py`)
The three distinct proxy-application functions — curl_cffi's `_new_curl_session` (serving + refresh legs), the shared `aiohttp_session_kwargs` (OAuth-login + ollama_cloud + model_context_sync), and botocore's `_get_boto3_client` (bedrock) — all apply the proxy unconditionally, destination-blind. The bedrock case drives with a loopback `endpoint_url` to pin construction-layer destination-independence at the one site that takes a destination. All three sites were falsified live: deleting the mapping breaks the test.

> Layer limit: R3 pins the construction layer only. A future bypass added at request time on these sessions, or via a session-level `CURLOPT_NOPROXY` with a matching host pattern, would leave construction-layer kwargs unchanged and this module green. Request-time and session-level bypass additions are caught by T-E3–T-E5's per-transport containment slices.

**R4 — Inventory guard** (`tests/test_egress_coverage.py`, extended)
A new `TestProxyApplicationSitesInventory` class fails on a silently-deleted proxy-application site (the six sites covered by R3) and fails on a `should_bypass` reference added to any custom-transport file. Both falsified live.

## Design update

`.system_design/TEST_SUITE.md` §5.4 ("Where I3 is proven") now records the two newly-proven rows.

## Test verification

- 26 tests green in the touched files (9 new + 17 in `test_egress_coverage.py`)
- 213 tests green across adjacent surfaces (egress, layer markers, wire-shape)
- All four claims empirically falsified (mutation caught on each)

## Scope and dependencies

- Depends on T-B1, T-B2, T-B3 (recorders) and T-E2 (SealedNetwork + bridge aiohttp slice) — all Done
- Does not depend on T-E3, T-E4, T-E5 (per-transport containment slices); those are the construction layer's complement
- Does not address T-E6 (per-entry-point fail-closed coverage) — explicitly out of scope


shelpuk:
Addressed — pushed as commit `4040413`. The fixture's docstring now reads ``asyncio.run`` (line 220), matching the current ``src/kitty/bridge_runner.py`` where ``asyncio.run(run())`` sits at line 220 inside ``main()``'s ``with contextlib.suppress(KeyboardInterrupt):`` block.

shelpuk:
Addressed the `CurlOpt` import-path suggestion — pushed as commit `a888e1e`. The module now imports `from curl_cffi import CurlOpt`, matching `tests/test_curl_cffi_transport_contract.py:41`, `tests/harness/test_curl_cffi_containment_slice.py:86`, and `src/kitty/providers/openai_subscription.py:44`. Re-ran `tests/test_egress_asymmetry.py`: 4 passed.

(The earlier `asyncio.run` line-reference comment was the same one already fixed and pushed as `4040413`; the review ran against a snapshot taken before that push landed.)

PR #206, opened 2026-09-17
KBR-257: the Gemini translator carries the upstream tool-call id on the response direction

shelpuk:
## Context (Product Owner)

When a Gemini-CLI conversation calls a tool through the bridge, the bridge used to strip the tool-call id on the way **back** to the Gemini client: every `functionCall` part arrived without the upstream provider's id, so the client's tool result had nothing exact to pair against. Most upstream providers tolerated the resulting name+position fallback; stricter ones would not, and the wire id the upstream set — a piece of the agent's conversation state — was silently lost at the bridge's response boundary.

This PR restores the id end-to-end. The upstream Chat Completions provider's `tool_calls[].id` now reaches the Gemini client as `functionCall.id`, so the client can echo it back on the next turn and the call/result pair is matched exactly. This is the response-direction mirror of PR #157 (KBR-195), which fixed the request side; with both sides landed, the standard Gemini agentic loop no longer depends on the name+position fallback.

The register row that would mechanically record the new delta on the response path is out of scope by design — it waits on T-D10 (KBR-59, To Do), the response-direction Reply projection, and the deferral is tracked in the spec and design doc so it cannot be forgotten.

## What changed

- `src/kitty/bridge/gemini/translator.py` — `translate_response` and `translate_stream_chunk` now read the upstream `tc["id"]` and carry it onto the emitted `functionCall` part; the `id` key is omitted when the upstream sent none. The streaming open branch stores the id in `_tool_call_meta` (now matching its long-standing `{index: {id, name}}` comment); the emit reads it with `.get("id")` so the `{"name": "unknown"}` fallback stays safe.
- `tests/test_gemini_translator.py` — new `TestTranslateResponseToolCallIdEcho` (five cases, styled after KBR-195's request-side class): sync present/absent echo, streaming carry across chunks, per-index carry on a two-call stream, `reset()` isolation. Two of the three original cases were red at the base revision before the fix (KBR-221 plan §16 discipline); the absent-sync case is the regression guard.
- `tests/bridge/test_messages_wire_translated_streams.py` — the pre-existing Gemini server-level assertion pinned the bug by exact equality (`calls == [{"name": "Read", "args": {"path": "a"}}]`); updated to the new correct end-to-end shape, where the Anthropic-native upstream's `toolu_1` reaches the Gemini client across all five Messages-wire adapters.
- `.system_design/SYSTEM_DESIGN.md` §4 X4 — records the carry rule and its why: the `is not None` vs `or` presence-test asymmetry (load-bearing for the empty-id round trip), the name-keyed streaming open, and the T-D10 / KBR-59 tracking link with the `reply.parts[*].id` anchor.

## Out of scope (tracked)

- The register row claiming `reply.parts[*].id` — waits on T-D10 ([KBR-59](https://shelpuk.atlassian.net/browse/KBR-59), To Do); the row to add there is the response-direction mirror of M18/M19.
- No Gemini reader work — the reader already handles `functionCall.id` (KBR-36).

## Verification

- Full focused sweep green: 347 passed (`tests/test_gemini_translator.py`, Gemini readers, `tests/bridge/test_messages_wire_translated_streams.py`).
- Red/green split independently reproduced at the base revision by the reviewer: 2 failed with a legible `assert None == 'call_1'`, 1 passed.
- `ruff check` clean on all touched files; `mypy src/kitty` clean (94 files).
- Reviews: `system-design-reviewer` (1 blocker — the server-level assertion — plus should-fixes, all applied) and `code-reviewer` (APPROVE).

[KBR-59]: https://shelpuk.atlassian.net/browse/KBR-59?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


shelpuk:
Addressed both findings from the round-1 automated review (commit c27e247):

**Warning — `test_reset_clears_ids_between_requests` could not fail on the regression it names.** Agreed, and traced to the root: the finish emit iterates `sorted(self._tool_call_buffers)`, not `_tool_call_meta`; a buffer for an index is created only inside the same open branch that overwrites the meta entry; and argument-only chunks for a bufferless index are dropped by the `idx in self._tool_call_buffers` guard. A stale meta entry is therefore unreachable from any emit — meta-clearing on `reset()` is defense-in-depth with no observable shadow. Rather than leave a test that passes either way (or pin the `_tool_call_meta` dict directly, coupling the test to internals), the test is **deleted**. The multi-call and streaming-carry tests stay — both are falsifiable.

**Suggestion — one-line docstrings on `translate_response` / `translate_stream_chunk`.** Both methods now carry full Google-style `Args:`/`Returns:` docstrings documenting the id-carry rule, the name-keyed open (per §4.2), and the pointer to `SYSTEM_DESIGN.md` §4 X4 for the `is not None` vs `or` round-trip asymmetry — so the rule is documented where the next reader of either method will look first.

Verification: focused sweep 346 passed (one fewer — the deleted test); ruff clean; mypy src/kitty clean.

PR #207, opened 2026-09-17
KBR-158: drop the dead Profile.base_url field; reject top-level base_url with a pointed message

shelpuk:

A profile today carries its base URL in **`provider_config["base_url"]`** — the wizards write it there, the validation pass and five provider adapters read it from there. A second, typed field — `Profile.base_url` — has lived on the schema the whole time and **nothing has ever read it**. The two channels validate differently: the live one permits `http://` for local vLLM/LM Studio/Ollama endpoints; the dead one would only accept `https://`, capped at 2083 characters, pydantic-normalised. Worse, pydantic's default `extra='ignore'` means a hand-written profile carrying top-level `base_url` was **silently dropped** — the request went to the provider's default endpoint, not the one the user typed. That is the precise failure mode the ticket exists to eliminate, now applied to every user.

This change makes `provider_config["base_url"]` the single authoritative channel, deletes the dead field, and rejects any profile that sets top-level `base_url` with a message pointing at the live channel. A user who set a URL in the wrong place now gets a clear error instead of a misrouted request.

The rejection is **value-aware** (rejects non-`None`), not key-presence — because `store.py` serialises every profile with `"base_url": null` from the deleted optional field, and `store._deserialize_entry`'s existing invalid-entry handling would otherwise silently drop every existing user's profile list.

## What changed

- **`src/kitty/profiles/schema.py`**: dropped `Profile.base_url`, the `HttpsUrl` class, and the `UrlConstraints`/`HttpUrl` imports only it used; added a `@model_validator(mode='before')` that raises `ValueError` (pydantic wraps in `ValidationError`) when raw input carries a non-`None` top-level `base_url`. Guarded by `isinstance(data, dict)` because `mode='before'` also receives model instances.
- **`tests/test_profile_schema.py`**: rewrote the five dead-field cases. New: `TestProfileBaseUrlChannel` (the channel is `provider_config["base_url"]`; both `http://` and `https://` pass through verbatim), `TestProfileRejectsTopLevelBaseUrl` (rejection + explicit `None` + legacy-`null` migration pin), and a round-trip test pinning that the dump no longer carries `base_url`.
- **`tests/harness/bridge.py`**: `profile_for` docstring updated to the post-deletion state.
- **`src/kitty/launchers/base.py`**: `build_spawn_config` docstring updated; the profile carries the URL in `provider_config`, not as a top-level field.
- **`.system_design/SYSTEM_DESIGN.md`**: new §10 "Profile schema: one base-URL channel" with the D9 decision, the rejected alternatives (make-authoritative-and-migrate; type `provider_config`; silent deletion), the value-aware rationale, and the store-path note.
- **`.system_design/TEST_SUITE.md`** §7.5.2: the `Profile.base_url` sentence replaced.
- **`.system_design/steps/kbr158_delete_dead_profile_base_url.md`**: canonical step record (`scripts/regenerate_step_index.py` does not exist in the repo, so no index regeneration runs — recorded against the standing memory note).

## Where the rejection reaches whom — and where it does not

The rejection fires where a `Profile` is constructed in code. A hand-edited *store file* with a non-`null` top-level `base_url` is dropped by `store._deserialize_entry`'s existing invalid-entry handling (warning log + skipped entry) — the F44 contract the store already applies to any invalid entry. Surfacing per-entry validation messages from the store is out of scope; the next reader has the note in SYSTEM_DESIGN.md §10.3 and the step file.

## Verification

- `pytest tests/test_profile_schema.py`: 45 passed (was 36 before; +9 net).
- Targeted sweep covering every direct consumer (schema + resolver + balancing + CLI + integration + every provider test): **1180 passed in 6.37s**.
- Full `pytest -m "l1 or l2"` (the CI gate) could not complete locally due to the recorded L1-clean-tests-hang-under-contention behaviour with a parallel-session sweep in another worktree; CI's per-PR run is the uncontended gate and will exercise it.

## Test plan for the reviewer

L1 only. No L2/L3/L4 test added — per TEST_SUITE.md's layer model and the reviewer's note, the "one channel carries the base URL" half of the acceptance clause is pinned by `assert "base_url" not in Profile.model_fields` (field absence *is* the pin; no resolver can read what does not exist), and the "rejected with a message" half is pinned by the four rejection tests with regex assertions against `provider_config["base_url"]`.


shelpuk:
**CI fix pushed (`10d9be3`) — all six matrix jobs failed on the same single test.**

Root cause: `tests/harness/test_bridge.py::TestTheProfileFactory::test_a_profile_carries_the_binding_and_is_schema_valid` asserted `profile.base_url is None` — the deleted field's old "not set" shape — as a negative pin that the harness profile factory does not set the dead channel. Post-deletion the attribute no longer exists, so the line raised `AttributeError` on the CI merge ref (`1 failed, 7915 passed` in each job — the same one test everywhere).

The line was redundant, not load-bearing:
- the positive pin on the line above it asserts the live channel (`profile.provider_config["base_url"] == recorder.base_url`) — that stays;
- the field-absence invariant is pinned by L1 in `tests/test_profile_schema.py::TestProfileBaseUrlChannel::test_base_url_is_not_a_declared_field`;
- the next-reader note lives on the `profile_for` docstring in `tests/harness/bridge.py`.

Swept the whole `tests/` tree afterwards for any other `profile.base_url` / `.base_url is None` / `hasattr(..., "base_url")` pattern — no other consumer of the deleted attribute exists. Local verification before push: `tests/harness/` + Profile neighbours — **2425 passed, 2 skipped** (the earlier targeted sweep missed `tests/harness/test_bridge.py`; that is the miss this fix corrects).

PR #214, opened 2026-09-17
KBR-84 (T-G8): aiohttp proxy behaviour contract

shelpuk:

Kitty Bridge keeps AI-provider traffic contained by routing it through the user's configured egress proxy. On the aiohttp path — which serves **every** standard provider route — that containment is set once, at session construction (`_build_client_session`), so no individual call site can forget it. This PR converts that design's *assumption* about aiohttp's behaviour into a **checked fact**: a dependency contract that fails loudly if a future aiohttp upgrade (within the declared `>=3.11,<3.14` range) silently changes the semantics containment relies on.

No product code changes. One new test file, one allowlist registration.

## What was implemented

`tests/test_aiohttp_transport_contract.py` — the aiohttp twin of KBR-161's curl_cffi contract (`tests/test_curl_cffi_transport_contract.py`) and KBR-64's botocore contract (`tests/harness/test_botocore_transport_contract.py`), all under TEST_SUITE.md §6.2.4 ("Dependency behaviour contracts"). Marked `l2` (fast gate).

Nine cases, every one against real local sockets (an aiohttp target app and recording proxy gateways, ports 8791-8793 — no collision with the siblings' 8781/8782):

| Req | Case |
|---|---|
| R1 | session `proxy=` honoured — gateway hit once, target zero |
| R2 | session `proxy_auth=` reaches the gateway as `Proxy-Authorization` (base64-derived expected value) |
| R3 | **headline**: a per-request `proxy=None` cannot escape the session-level proxy — `_session_for`'s two-session containment design rests on aiohttp's resolver (`client.py:575-576`: `if proxy is None: proxy = self._default_proxy`) |
| R4 | the opposite branch pinned too: a per-request explicit `proxy=` overrides the session-level one, so a future reversal turns red rather than silently altering containment |
| R5 | ambient `HTTP_PROXY`/`http_proxy` is ignored by default (`trust_env=False`) — a user's shell cannot defeat the configured egress; dead-address falsification channel |
| R6 | the adapter applies the contract: `_build_client_session(proxied=True)` passes `proxy=`/`proxy_auth=`; `proxied=False` leaks neither |
| — | self-guard: the module counts its own behavioural cases (floor = current count), so silent deletion of any one case is a red |

Plus the one-entry registration on `tests/test_layer_selection.py`'s L2 contract allowlist, following the botocore twin's comment pattern.

## Testing

- New file: 9/9 pass on aiohttp 3.13.5 (`pytest tests/test_aiohttp_transport_contract.py`), also green with `-W error::ResourceWarning` (the R6 spy stubs the `TCPConnector` so no connector is leaked).
- `pytest -m l2 -q`: 923 passed, 2 skipped.
- Full suite: 8001 passed, 13 skipped, 0 failures (25 min).
- `ruff check` + `ruff format --check` clean.
- Requirements: `.requirements/20260917T120456Z_kbr84_t_g8_aiohttp_proxy_contract/REQUIREMENTS.md` (R1-R6, AC1-AC8, testing plan).

## Deliberately not done

- **No `src/kitty/**` changes.** The contract passes against current code — it pins behaviour that already holds; its job is to turn red if a bump inside the pin range breaks it.
- **Does not use T-W5's TLS harness** (`harness/connect_proxy`): these claims are about plain-HTTP proxy semantics, not TLS-in-TLS, and the curl_cffi sibling's two-local-app pattern fits. The TLS stack remains §7.3's concern.

Closes [KBR-84](https://shelpuk.atlassian.net/browse/KBR-84) on merge.

[KBR-84]: https://shelpuk.atlassian.net/browse/KBR-84?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


shelpuk:
**Round 1 CI post-mortem — root cause + fix pushed.**

All six matrix legs failed with exactly one test:

```
FAILED tests/test_aiohttp_transport_contract.py::test_the_module_finds_known_positives
ModuleNotFoundError: No module named 'tests'
```

**Root cause.** The self-guard re-imported its own module via
`import tests.test_aiohttp_transport_contract as mod`. `tests/` is a pytest
rootdir without an `__init__.py`, so that import only resolves when the
repository root is on `sys.path`. Locally I drive pytest as
`python -m pytest`, which prepends the cwd (the repo root) — so it passed
every local run. The CI fast-gate job calls bare `pytest`
(`.github/workflows/tests.yml:110`), which does not. Classic
works-on-my-machine, and the falsification channel for it (running the
CI-style invocation locally) is now part of the workflow I follow for
contract modules.

**Fix (6ea0de0).** The guard walks `sys.modules[__name__].__dict__`
instead of re-importing — same shape, no import-path dependency, works
under both `python -m pytest` and bare `pytest` from any cwd. Verified
against the CI-style invocation locally before pushing.

No product code touched; no other test affected (7988/7988 passed on
every leg; the six legs share the one failure).

shelpuk:
**Round-2 §8.5 review addressed (commit 5aaef5e).**

The reviewer's note was correct: the floor of 7 counts *methods*, but
`test_ambient_http_proxy_is_ignored_by_default` is one method carrying
two parametrize arms. Deleting one arm (e.g. dropping `http_proxy`)
leaves the method count at 7 and the floor keeps passing while a full
direction of the R5 ambient-env claim silently disappears — the opposite
of what the guard's docstring claimed.

Adopted the reviewer's stronger fix (option 1): added a set-equality
assertion on the parametrize arms, mirroring the repo's dominant
self-guard shape (`test_upstream_url_single_rule.py`,
`test_answered_questions_are_settled.py`) — pin the known set rather
than rely on a count alone. The guard now covers both ways a behavioural
case can silently disappear. Docstring rewritten to match.

Verified under both `python -m pytest` and CI-style bare `pytest` before
push. Re-arming the CI monitor.

PR #216, opened 2026-09-17
T-F6 (KBR-75): the translator semantic property, via the projections

shelpuk:

Kitty Bridge's job, at its core, is to sit invisibly between a coding agent and an AI provider: the agent speaks Anthropic's "Messages" protocol, many providers speak OpenAI's "Chat Completions" dialect, and the bridge translates between them on every request. If that translation ever *loses or alters part of the agent's conversation* — a tool call that arrives without its arguments, a system prompt that silently disappears, a user message that gets truncated — the model sees a different conversation than the agent sent, and the failure shows up downstream as confused or broken agent behaviour with no obvious cause.

Today the translator is covered by ~100 example-based tests, each pinning one named conversion. That is the right shape for "does the system prompt lift correctly", but it cannot state the *general* rule the product actually needs: **for every conversation the agent can send, the translation preserves what the agent said.** This PR adds exactly that rule, as a property-based test: the `hypothesis` framework generates hundreds of realistic conversations (text, images, documents, tool calls, tool results, at any length) and asserts, for each one, that translating it Anthropic→Chat-Completions preserves the same semantic content — compared through the test suite's *independent* wire-format readers, never by asking the translator to check itself (a self-checking translator is the one trap the design doc §3.3.1 spends a full section on: it proves self-consistency, not fidelity).

The property also ships with its own falsification controls, per the plan's harness rule (§1.4): two deliberately-broken translator variants (one drops a wire flag, one drops a token budget) prove the property actually *detects* that class of defect rather than passing vacuously, and a "drops nothing" control keeps those detection checks from silently rotting.

Two things the property deliberately does **not** do, recorded so nobody re-litigates them:

- It does not cover fields the shared test-strategy library doesn't generate yet (`system`, `tool_choice`, `metadata`, `stop_sequences`, `top_k`, `effort`, `output_config`, `thinking`, cache breakpoints). Those have their own example-based coverage today; extending the strategy library is follow-up substrate work, not this ticket.
- It documents (rather than papers over) three *deliberate* translation behaviours that make the two wire projections structurally differ — document blocks riding an internal channel to Anthropic-family providers, the CC reader's hardcoded `type: "function"` on wrapped tools, and text-block joining — each with its design-doc citation, applied in exactly one place.

No product code changes. `src/kitty` untouched.

## What changed

| File | Change |
|---|---|
| `tests/bridge/test_messages_translator_property.py` | New (L1) — the semantic property (200 Hypothesis examples per run), the three named exemptions in one comparison helper, two falsification cases + negative control, and a strip-set derivation check tying the test's internal-key strip to the production registry |

## How it was built (TDD, review flow)

1. Requirements drafted → `system-design-reviewer` pass (10 findings incl. 2 blockers — the `ToolDecl.type` asymmetry and a wrong falsification channel — both applied before implementation) → Hypothesis then surfaced a third genuine asymmetry (text-block joining) on its first red run, folded into the same named-exemptions helper.
2. Watched every new test fail for the right reason before it passed; the two falsification cases and the negative control are committed red/green pairs, not assertions added after the fact.
3. `code-reviewer` pass on the finished file (this PR's review thread).

## Verification

- 4/4 new tests pass under `CI=1` (deterministic Hypothesis profile: derandomised, no deadline, no example database — the Windows/macOS Fast-gate legs will not flake on it).
- Full L1 gate green (`pytest -m l1 -q`, exit 0), layer-marker guard 51/51, `ruff` clean, `mypy src/kitty` clean (94 files), `lint-imports` 5 kept / 0 broken.

Closes KBR-75 (T-F6). Completes Epic F (component and property tests, L1).


shelpuk:
Round-1 review response — all four findings applied in 56d8174:

**S2 (negative control, dropped).** Confirmed: `_DropsNothing` overrode nothing, so its output is bit-identical to `MessagesTranslator()` and its two assertions duplicated the healthy check immediately above. Deleted the subclass and both control-assert blocks; each falsification test keeps the healthy-pass + defective-fail pair only. The module docstring's §1.4 paragraph no longer advertises a negative control.

**W2 (tautological strip mirror, fixed by hand-pinning).** Confirmed the tautology: the constant was *defined* as `ProviderAdapter._INTERNAL_KEYS - {...}` and the test recomputed the same expression. The constant keeps the derived definition (single source, no drift at the point of use), but the test now pins the expected 15 names literally, so any `_INTERNAL_KEYS` change fails it and forces a conscious look at the strip. Docstring rewritten to state the test's honest job.

**S1 (docstring, corrected).** Empirically confirmed: `_typed_leaf`'s default for an absent key is `None` (reader_chat_completions.py `default: Any = None` → `source.get(key, default)`), so the `_DropsStream` docstring now says the delta is `envelope.stream == None` against the inbound `bool`. The requirements doc's T2 paragraph was corrected the same way.

**W1 (CC-reader independence, enforced).** Confirmed the gap: `test_reader_anthropic_messages.py` has `test_the_reader_imports_nothing_from_kitty` (R1.3), but `test_reader_chat_completions.py` did not — even though every other harness module the oracle depends on (recorder, cache breakpoints, register, provider_aiohttp, eval_harness, contract) carries the same guard. Added `TestProtocolConformance` to `tests/harness/test_reader_chat_completions.py` mirroring the Anthropic reader's class: Projection-protocol check, `wire_format` declaration, and the `_KITTY_IMPORT`-regex independence guard.

Post-change local gates: 7/7 new and modified tests pass under `CI=1`, layer-marker guard 51/51, ruff clean, mypy `src/kitty` clean (94 files), lint-imports 5/0.

shelpuk:
Round-2 review response — the single warning applied in 7f2047c:

**W3 (`_turn_signature` lossy in undeclared ways, fixed by role + mixed-turn order).** Confirmed both losses by tracing the reduction:

- `Turn.role` was dropped by the `(non-text parts, joined-text)` reduction — a translator that flipped a turn's role would have passed silently on the strategy's text-only turns. The signature now carries `turn.role`.
- Cross-kind part interleaving was also lost — the old reduction split text and non-text parts into two buckets and lost their relative order, so a translator that hoisted text first in a mixed turn would have passed. The signature now branches: **text-only turns** reduce to `(role, joined_text)` (the M2/KBR-222 join clause covers per-block presence); **mixed turns** compare the full part tuple in wire order, catching ``[Image, Text]` → `[Text, Image]`` reordering. Neither translator nor reader reorders parts, so per-part equality is the honest comparison there.

**Two new falsifications, per §1.4** (a defence without a wired detection is a decorative assertion):

- `_SwapsTextOnlyUserRoles` — flips `user` → `assistant` on string-content messages only. A broader swap (assistant → user on a body with `tool_calls`) produces a body the CC reader rejects with `UnreadableBodyError`, which would have failed the falsification with the wrong signal rather than the property's assertion — the narrower mutation keeps the falsification honest. Uses `assume()` to focus on bodies with at least one text-only user turn.
- `_HoistsTextPartsFirst` — reorders mixed content lists to put text parts first. Uses `assume()` to focus on bodies where the hoist actually changes something (a body whose mixed lists are already text-first is unchanged by the hoist — Hypothesis caught that on the first red run and shrank to it). The hoist filter is ~88% rejection (mixed lists are rare in the strategy's space), so the test carries `suppress_health_check=[HealthCheck.filter_too_much]` with a comment naming that suppression as honest: mixed turns are exactly the inputs where the hoist could fire, and every filtered body was a vacuous run regardless.

**Module docstring updated** to declare both non-exemptions explicitly (`role` on every signature; mixed-turn full-tuple equality) so a future reader extending the strategy does not mistake either for an undeclared exemption.

**First red run produced a real bug.** The property itself still passed on the strategy's space (all 200 examples green before the new falsifications were added), but adding the role/hoist falsifications immediately surfaced the two defects the reviewer had predicted — the property genuinely was blind to both.

Post-change local gates: 6/6 property tests + 3/3 CC reader protocol tests pass under `CI=1`; layer-marker guard 51/51; ruff clean; mypy `src/kitty` clean (94 files); lint-imports 5 kept / 0 broken.

shelpuk:
Round-3 review response — the single suggestion applied in 4723cf2:

**S3 (`_SwapsTextOnlyUserRoles` docstring's failure mechanism, corrected).** Confirmed empirically with an out-of-tree probe before rewriting: a broad role swap (assistant → user on a message carrying `tool_calls`) does **not** raise `UnreadableBodyError`. The CC reader's user branch reads `content` against `_USER_PART_TYPES` and residualises everything else — its `_residualise` set is `{role, content, name}`, so `tool_calls` lands in the residual, `read_request` returns successfully, and the falsification dies later in `verify_total` with `ResidualFieldsError`. The probe output:

```
translated roles: [('user', 'str'), ('assistant', 'NoneType'), ('tool', 'str')]
swapped roles: ['assistant', 'user', 'tool']
read_request: OK — residual: {'messages[1].tool_calls': [...]}
verify_total raised: ResidualFieldsError -> reader could not classify ['messages[1].tool_calls'] ...
```

The docstring now states the true mechanism and the real reason for the narrowing: the narrow mutation keeps the falsification's failure signal pointed at the **role comparison** — the defence the test exists to validate — instead of letting it die earlier on reader bookkeeping. (The reviewer's parenthetical that a broader swap "would in fact be caught by the role comparison" is true only in the counterfactual where the residual were empty; in the actual reader the broad swap fails at `verify_total` before the comparison runs, which is exactly why the narrow form is the right instrument.)

No behavioural change — docstring only. Local gates re-verified: 6/6 property tests pass under `CI=1`; ruff, `mypy src/kitty`, `lint-imports` clean.

shelpuk:
Round-4 review response — the single finding applied in 37ba4c6:

**S4 (`_SwapsTextOnlyUserRoles` docstring's stated reason, corrected again — this time with the probe-verified truth).** Confirmed: my round-3 rewrite replaced the original `UnreadableBodyError` claim with a `ResidualFieldsError` claim that was equally wrong, because this falsification never calls `verify_total` — so neither reviewer's proposed mechanism matched the falsification's behaviour.

Probe-verified the broader-swap failure mode per body shape before rewriting:

| Body shape | Broader swap's failure mode |
|---|---|
| image-bearing user turn | `UnreadableBodyError` (assistant branch rejects `image_url` parts — `_ASSISTANT_PART_TYPES` is `{text, refusal}`). Role defence never fires. |
| tool_use assistant, flip only user→assistant | read OK, no residual — role defence catches it. |
| tool_use assistant, full user↔assistant flip | read OK with residualised `tool_calls`, but the falsification doesn't call `verify_total` — the comparison runs and the role defence catches it. |
| string-content user turn | read OK, no residual — role defence catches it. |

**The narrow mutation's actual reason is the image case alone** — the only case where a broader swap fails with a reader error instead of the role defence. The tool_use case is caught either way (as the round-3 reviewer said), so the narrow mutation is not required for it. The docstring now states this probe-verified reason.

No behavioural change — docstring only. Local gates re-verified: 6/6 property tests pass under `CI=1`; ruff, `mypy src/kitty`, `lint-imports` clean.

shelpuk:
Round-5 review response — the single finding applied in b2ac66e:

**S5 (`_SwapsTextOnlyUserRoles` docstring conflated two broader-swap scenarios, simplified).** Confirmed: my round-4 rewrite kept the correct conclusion ("the image case is the one that forces the narrow mutation") but used a two-bullet structure that mixed the image case with a different broader swap (one that flips assistant → user on tool_use assistant messages). The reviewer traced both scenarios and concluded the image case is what forces the narrowing; the tool_use bullet was an over-elaboration that conflated two separate hypotheticals.

Simpler and more accurate: drop the tool_use bullet entirely, keep the one concrete case (the image one), and let the conclusion stand on its own. One paragraph, one reader-side failure mode, one conclusion.

No behavioural change — docstring only. 6/6 property tests still pass under `CI=1`; ruff, `mypy src/kitty`, `lint-imports` clean.

shelpuk:
Round-6 review response — the reviewer formally returned **no findings** (by their protocol, round-6 withholds suggestion-level findings), but their review body identified a genuine docstring inaccuracy that no prior round caught: `_conversations_equivalent`'s own docstring (lines 165-170) still described the third exemption's reduction as `(non-text parts, joined text)` — the shape the round-2 fix removed in favour of a branched reduction carrying `Turn.role`. The inline comment at lines 206-218 and the module docstring at lines 45-50 are both correct; only this function-level docstring was stale.

Applied in 30ea3cb — even though the finding was withheld at round 6, a stale docstring is a relevant defect and was in the reviewer's prose anyway. The bullet is rewritten to describe what the implementation now does:

- Text-only turns: `(role, joined_text)` on both sides (the M2 text-join exemption).
- `Turn.role` rides on every signature (not exempted, round-2 fix).
- Mixed turns: `(role, tuple(turn.parts))` for full wire-order equality (not exempted, round-2 fix).

No behavioural change — docstring only. 6/6 property tests still pass under `CI=1`; ruff, `mypy src/kitty`, `lint-imports` clean.

shelpuk:
Round-7 review response — the warning applied in 096564d:

**W4 (CI profile's `too_slow` suppression silently dropped on the hoist test, fixed).** Confirmed the settings-inheritance interaction: a `@settings(suppress_health_check=...)` on a test **replaces** the parent profile's list rather than extending it. The hoist falsification's override — `suppress_health_check=[HealthCheck.filter_too_much]` — therefore dropped the CI profile's `[HealthCheck.too_slow]` on exactly the test whose 88% filter rate makes that suppression load-bearing (slow generation on Windows/macOS legs trips `too_slow` once enough filtered examples accumulate generation time).

Fixed by naming both health checks explicitly in the test-level override:

```python
@settings(
    max_examples=200,
    suppress_health_check=[HealthCheck.filter_too_much, HealthCheck.too_slow],
)
```

The block comment now also names the interaction, so a future reader does not "simplify" the list back down and silently re-drop the profile's `too_slow`.

The withheld round-7 suggestion (documenting the interaction) is folded into the same comment.

No behavioural change under `CI=1` (the profile's `too_slow` is restored to its intended effect). 6/6 property tests still pass; ruff, `mypy src/kitty`, `lint-imports` clean.

shelpuk:
Round-8 review response — the reviewer formally returned **"No findings"** (their strictest protocol: warnings/criticals only, suggestion-level findings out of scope since round 6). The body verified every prior fix (rounds 1–7) on the head commit `096564d`.

This is the convergence point the workflow is waiting for: **the reviewer has stopped providing actionable feedback**. Per the task's working contract, I'm not merging — the reviewer / product-owner does.

Final PR state on the head commit:
- One new test file (`tests/bridge/test_messages_translator_property.py`, 565 lines).
- One 38-line `TestProtocolConformance` class in `tests/harness/test_reader_chat_completions.py` (round-1 fix, mirrors the Anthropic reader's class and reuses `_KITTY_IMPORT`).
- 8 review rounds (5 rounds of automated findings addressed; 3 of clean / prose-only acknowledgement).
- Local gates green throughout: ruff, `mypy src/kitty` (94 files), `lint-imports` (5 contracts kept / 0 broken), layer-marker guard (51/51), 6/6 property tests + 3/3 CC reader protocol tests pass under `CI=1`.
- CI state: seven checks green in run 8; six test-matrix legs + `ci-required` pending. The CI watch is live and will fire when those land.

KBR-75's Jira status stays **In Progress** until the PR merges — by you.

PR #224, opened 2026-09-17
KBR-279: align Ollama request reader on the strict function.name check

shelpuk:

The comparison oracle that validates kitty's vendor adapters reads every
provider's wire traffic through per-format "readers". One rule those readers
must agree on: a tool call without a usable name cannot be trusted — it cannot
be paired with its result or addressed downstream. The Ollama reader's
**reply** direction has enforced that rule since KBR-267; its **request**
direction still silently invented an empty-string name for malformed input.
This PR closes that gap so both directions of the Ollama reader enforce the
same rule from one shared implementation. Real Ollama traffic is unaffected —
every published Ollama example carries a valid tool name — so this is a
malformed-input hardening with zero behaviour change on the happy path.

## What changed

* **One shared rule, one place.** New `_require_tool_call_name` helper in
  `tests/harness/reader_ollama.py`; both the request reader's
  `_read_tool_calls` and the reply projection's pre-check delegate to it
  (§7.4.1's within-module anti-drift rule). The helper's docstring records two
  load-bearing invariants: the raise-ordering contract with the non-object
  `function` guard, and the cross-reader residual-set delta T-D8 should not
  read as drift.
* **Three falsifications + control** in `tests/harness/test_reader_ollama.py::TestNameRequired`,
  mirroring the reply side's class in `test_reader_ollama_reply.py`:
  missing / empty / non-string `function.name` each raise
  `UnreadableBodyError`; the control projects the published
  no-streaming-with-tools multi-turn request cleanly.
* **Design-doc truth.** `TEST_SUITE.md` §7.4.2 rule 7 row 2's Because cell now
  records the per-reader picture: the Responses reader implements the row's
  general residualise+project prescription verbatim; Anthropic, Chat
  Completions (non-string), Gemini (non-string) and Ollama (absent / empty /
  non-string, both directions) deliberately raise. The row's Outcome cell is
  unchanged — an implementation-time check found the earlier
  "prescription is dead letter" claim wrong (it missed Responses).

## Testing

* TDD: the three falsifications were written first and failed with
  "DID NOT RAISE" (the exact defect); the control passed before and after.
* Full `tests/harness/test_reader_ollama.py` + `test_reader_ollama_reply.py`
  green (128 tests); no other test file touches the Ollama readers.
* Gates: `ruff` ✓ · `lint-imports` ✓ · `mypy src/kitty` ✓ · full suite ✓
  (Linux).

## Review flow

* system-design-reviewer: two rounds on the REQUIREMENTS.md (8 findings,
  all folded in) + one confirmation round; one implementation-time factual
  correction to its row-2 proposal (Responses reader) recorded in the step
  file and PR body.
* code-reviewer: ran on the diff against the REQUIREMENTS.md acceptance
  criteria (see PR comments for the outcome).

Jira: [KBR-279](https://shelpuk.atlassian.net/browse/KBR-279) ·
Follow-up to KBR-267 (PR #215).

[KBR-279]: https://shelpuk.atlassian.net/browse/KBR-279?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


shelpuk:
CI fully green — all 14 checks pass, including `ci-required` and the Python 3.10–3.13 matrix on ubuntu/macOS/windows (macOS was the long pole at ~90 min). No reviewer comments yet. Awaiting human review; not merging.

PR #246, opened 2026-09-20
KBR-293: judge-first custom-transport holds on the /v1/responses and /v1/gemini segments

shelpuk:

KBR-287 taught us that when the bridge fails over to a provider driven by custom transport (Bedrock, Ollama Cloud, or the Codex subscription) and that provider returns an **empty completion**, the agent could receive a well-formed but contentless reply — the bridge never tried another backend, and a balancing pool kept routing to the broken one. KBR-287 fixed that on the Chat Completions route only. KBR-293 asked whether the same defect lives on the other two routes this product serves: **/v1/responses** (Codex CLI) and **/v1/gemini** (Gemini CLI).

**It did — and it was worse.** Bridge-level tests, watched failing before any fix, proved:

- On `/v1/responses`, empty completions from all three custom-transport providers sailed through as skeletons (13 of 14 falsification cases red pre-fix).
- On `/v1/gemini`, the problem was broader: the providers emit the *wrong wire format* for that route entirely, so even **content-bearing** completions never arrived as readable Gemini events (14 of 14 red).

This PR ports KBR-287's discipline — collect the provider's response, judge whether it carries content *before anything is sent*, retry via the existing empty-response ladder when it doesn't, and end in the route's honest error terminal when everything is empty — to both routes. It adds the one step KBR-287 never needed: translating the judged content into the route's own wire format, which is what makes content reachable on these routes at all. It also corrects three older tests whose fake upstreams mimicked a wire real providers never emit (the exact blind spot that let this class of bug hide).

**One deliberate product trade-off, flagged for sign-off:** `/v1/responses` with a Codex subscription used to pass the subscription's native stream straight through, reasoning summaries included. Post-fix it goes through judge-then-translate, and the parser drops reasoning summaries. A *reasoning-only* completion on that combination flips from "delivered" to "retried, then honest empty-error". Recorded in `SYSTEM_DESIGN.md` §5.4; the alternative (a special-case passthrough branch) was rejected because it would forfeit the empty-completion guard for exactly that cell. A future ticket can restore reasoning by widening the parser's projection — no branch fork needed.

**Also on the record:** the failure-after-bytes safety rule on these two branches is now satisfied *by construction* (nothing is sent before the verdict), so the KBR-247-era test pinning the old mechanism was rewritten to pin the stronger guarantee. The rule itself is unchanged everywhere content still streams incrementally. The `/v1/messages` route's custom segment has the same missing emptiness gate but no wire problem — recorded as a deferred, smaller-class follow-up.

## What changed

- `src/kitty/bridge/server.py` — `_stream_responses` and `_stream_gemini` custom-transport segments: collect → parse (adapter parser / Responses-SSE fallback) → synthesise the KBR-287 CC payload list → judge via the shared `_cc_chunk_carries_content` → content arm (route-translated write, lifecycle, log-on-release usage) / empty arm (KBR-287 class-agnostic ladder; custom stays custom, plain crosses via fall-through) / D4 exhaustion terminal (`code: "empty_response"` + `response.completed(incomplete)` on Responses; `reason: "empty_response"` + EOF on Gemini). Attempt bounds widened to `n_backends + len(_EMPTY_FINAL_DELAYS)` with the final-delay prologue; vacuous `_bytes_written` guards dropped.
- `tests/bridge/test_responses_custom_transport_empty_hold.py`, `tests/bridge/test_gemini_custom_transport_empty_hold.py` — new: the KBR-287 harness shape, parametrised over the three adapters, route-protocol-parsed content oracles (never raw substrings — the KBR-249 vacuous-oracle trap). Falsification baseline: 13/14 and 14/14 red pre-fix; green post-fix.
- `tests/bridge/test_balancing_server.py` — KBR-254 sibling stubs corrected to real Bedrock emission (`_fake_hello_cc_stream`); Responses-SSE stub and its now-false docstring (`_fake_hello_stream`) removed; one inline Responses-SSE stub corrected likewise.
- `tests/bridge/test_post_emission_no_failover.py` — Q14(a) test rewritten for the by-construction pre-emission contract (stronger guarantee pinned).
- `.system_design/SYSTEM_DESIGN.md` §5.4 — the KBR-293 paragraph: fix shape, per-route D4 terminals, translator-reuse state-safety argument, fidelity callout, recorded asymmetries (custom transport-error terminal ships no lifecycle close, pre-existing; /v1/messages no-gate, deferred).
- `.system_design/steps/kbr293_sibling_custom_transport_holds.md` — step file + implementation notes.

## Verification

- New tests: 29 green (15 responses + 14 gemini); KBR-287's 14 green (no CC-route regression); balancing file 59 green; post-emission file 21 green.
- `ruff` clean, `mypy` clean on `server.py`.
- Full suite green on the final state locally (round 3) — CI is the authoritative run.


shelpuk:
Thanks — addressed. The Responses D4 exhaust test now also pins the synthesized `response.completed` carrying `status: incomplete` that the arm emits before EOF (one-line addition at `test_responses_custom_transport_empty_hold.py:659`). All 15 tests in the file still green, full suite still clean.

shelpuk:
Round 3 (clean full-suite run, no mid-run edits) caught one real consequence outside the new tests: the streaming cells of `tests/bridge/test_responses_string_input.py::TestTheUpstreamBodyOnTheCustomTransport` were stubbing the recording provider's `stream_request` with a lone `response.completed` event — a pre-fix wire shape that pre-fix rode the raw passthrough to the client as a skeleton. Under the new judge-first segment it correctly parses as content-free and ladders to the D4 terminal (exactly the defect class this ticket closes), so the stub's own one-attempt assertion failed (3 attempts instead of 1).

Fixed the stub to write content-bearing Responses-SSE (`output_text.delta` + `response.completed`) — the same correction round 1 applied to the KBR-254 sibling stubs. The test's point is the shipped `_original_body`, not the empty-ladder shape. 44/44 in the file green; round-4 full suite in flight to confirm no other surprises.

shelpuk:
Addressed. The gemini D4 exhaust test now parses the `data:` payloads and asserts the error event's `error.reason` carries `empty_response` — replacing the raw substring check (the KBR-249 vacuous-oracle trap), mirroring the responses twin's parsed error-event oracle. 14/14 in the file green; pushed as 723bebe.

shelpuk:
Addressed. The README's empty-response ladder section now states the custom-transport backends are held on every route (Chat Completions, Responses, and Gemini clients), notes the route-native event conversion on `/v1/responses` and `/v1/gemini`, and replaces the singular-route wording in the exhaustion-terminal sentence. README guard tests (28) green; pushed as 89037d9.

shelpuk:
Acknowledged, not changing — KBR-287 parity. The CC twin's empty arm at server.py:8531 calls `_select_backend()` bare in the same position; my port mirrors that. Between the `_any_healthy_backend()` guard and the `_select_backend()` call there is no `await` in the same branch, so `AllBackendsUnhealthyError` cannot fire in that window (one synchronous check, one synchronous call, no coroutine switch). Adding the wrap here would (a) diverge from the twin's deliberate byte-for-byte shape (KBR-277 mirror convention) and (b) violate the "no error handling for impossible scenarios" rule. Worth recording: if the semantic gap the reviewer names (require-streaming / backup-tier holdout in the select path that the health check doesn't model) is real, it is a TWIN-wide latent issue, not a KBR-293 one, and the fix is a separate follow-up that lands the wrap on the CC twin and both siblings together. Pushed 89037d9 stands.

